Issue #117, opened 2026-02-02
ADR 006: Namespace prefix resolution is internal service behavior; fix name extraction bugs


jeffmccune:
## Summary

Organization and project names returned by the API are derived by parsing Kubernetes namespace names instead of reading from the canonical labels set during creation. This causes incorrect names when namespace prefixes are configured, producing artifacts like a project named `holos-o-holos` instead of `holos`.

Additionally, the `--namespace-prefix` flag defaults to `""` (empty), but should default to `"holos-"` so that all console-managed namespaces are clearly distinguished in the cluster.

This issue includes an ADR formalizing the namespace resolution design decisions.

### Status

Proposed

### Context

The console server manages Kubernetes Namespace objects on behalf of users. Each namespace name is constructed from configurable prefixes (`--namespace-prefix`, `--organization-prefix`, `--project-prefix`) combined with the user-facing resource name. For example, with defaults `holos-`, `o-`, and `p-`, an organization named `acme` maps to the namespace `holos-o-acme`.

The current implementation exposes prefix internals to callers in two ways: (1) `OrgFromNamespace`/`ProjectFromNamespace` use `strings.TrimPrefix` which silently returns garbage when a namespace name does not match the expected prefix, and (2) `buildOrganization` always parses namespace names instead of reading from labels, leaking the prefix scheme into API responses.

### Decisions

1. **Namespace prefix configuration is internal service behavior, not exposed to callers.** API consumers operate exclusively with logical resource names (e.g. `acme`, `my-project`). The mapping between logical names and Kubernetes namespace names is an implementation detail of the server. No prefix value or namespace name is ever returned in API responses.

2. **The resolver uses configured prefix values when constructing and parsing namespace names.** `OrgNamespace`/`ProjectNamespace` construct names by concatenating prefixes. `OrgFromNamespace`/`ProjectFromNamespace` parse names by validating and stripping the expected prefix. Labels (`console.holos.run/organization`, `console.holos.run/project`) remain the authoritative source for resource names and must be preferred by callers when available.

[...]

PR #118, opened 2026-02-02
Fix org/project name extraction to use labels instead of namespace parsing

jeffmccune:
## Summary
- Fix `buildOrganization` to read the `OrganizationLabel` from namespace labels instead of parsing the namespace name with `OrgFromNamespace`
- Add warning log to `buildProject` fallback path when `ProjectLabel` is missing
- Change `--namespace-prefix` default from `""` to `"holos-"` so console-managed namespaces are clearly distinguished
- Deprecate `OrgFromNamespace` and `ProjectFromNamespace` resolver methods
- Fix `orgNS` test helper to set `OrganizationLabel` matching production `CreateOrganization`
- Update all test fixtures to use `holos-` namespace prefix matching the new default
- Fix proto comments for `Project.name` and `CreateProjectRequest.name`

Closes: #117

## Test plan
- [x] `make generate` succeeds
- [x] `make test` passes (org, project, resolver, rbac, cli tests all green; only pre-existing `TestScripts/grpc_reflection` fails due to missing TLS certs)
- [x] New tests verify names come from labels, not namespace parsing
- [x] Label takes precedence when it differs from what `OrgFromNamespace` would produce
- [x] Default `--namespace-prefix` is `"holos-"`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -145,7 +145,7 @@ Namespace prefix scheme (three-part naming: `{namespace-prefix}{type-prefix}{nam
 - Organizations: `{namespace-prefix}{organization-prefix}{name}` (resource-type label: `organization`)
 - Projects: `{namespace-prefix}{project-prefix}{name}` (resource-type label: `project`, optional organization label for IAM inheritance, project label stores project name)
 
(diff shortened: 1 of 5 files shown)


jeffmccune:
Closing to refine the plan in #117, I do want it hard coded.  It needs to filter out names that don't match or error.

Issue #231, opened 2026-03-31
Migrate slow E2E tests to fast unit tests with ConnectRPC mocking


jeffmccune:
## Problem

The E2E test suite is slow and requires a full Kubernetes cluster for most tests. 4 of 6 E2E spec files depend on live K8s resources (orgs, projects, secrets), forcing serial execution, 30-second server startup overhead, and 2 retries in CI. This creates a slow feedback loop for developers.

## Acceptance Criteria

- [ ] ConnectRPC service clients can be mocked in Vitest unit tests using a documented, reusable pattern
- [ ] The mock approach works with `@connectrpc/connect-query` hooks (useQuery, useMutation)
- [ ] Tests that exercise pure UI behavior (rendering, interaction, navigation logic) are implemented as unit tests, not E2E tests
- [ ] The remaining E2E tests cover only behaviors that genuinely require a full stack (OIDC login flow, real K8s secret CRUD)
- [ ] `docs/frontend-patterns.md` (or a new `docs/testing.md`) documents the ConnectRPC mock pattern with a worked example
- [ ] AGENTS.md is updated to reference the new/updated doc and to state the preference: unit tests first, E2E only for full-stack behaviors
- [ ] `make test-ui` remains the fast feedback path; `make test-e2e` is reserved for integration confidence

## Background

ConnectRPC's TypeScript generated clients are stable and backwards-compatible. The protobuf schema in `proto/` and generated code in `frontend/src/gen/` change infrequently and only in additive ways. This makes it safe and practical to mock the ConnectRPC layer in unit tests.

Current mocking patterns in the codebase (vi.mock for Router, sidebar, auth context, toast) show the team already uses this approach for other dependencies. Extending it to ConnectRPC clients follows the same pattern.

## Current State

| Spec file | K8s required | Can be unit tests? |
|-----------|-------------|-------------------|
| `auth.spec.ts` | No | Partially — OIDC redirect flow needs E2E; token display can be unit tested (already is in `-profile.test.tsx`) |
| `about.spec.ts` | No | Yes — sidebar content and version card are pure rendering |
| `secrets-grid.spec.ts` | Yes | Yes — list rendering, sorting, navigation are pure UI given mocked RPC data |
| `navigation.spec.ts` | Yes | Mostly — org/project picker rendering and nav logic; deep-link navigation needs E2E |
| `copy-toast.spec.ts` | Yes | Yes — clipboard + toast behavior is fully testable with mocked data |
[...]

PR #233, opened 2026-03-31
Migrate slow E2E tests to fast unit tests with ConnectRPC mocking

jeffmccune:
## Summary

- Add `docs/testing.md` with a documented ConnectRPC mock pattern, decision rule table (unit vs E2E), route-file naming convention, and test file index
- Migrate `about.spec.ts`, `copy-toast.spec.ts`, and `secrets-grid.spec.ts` to unit tests and delete them
- Reduce `navigation.spec.ts` from 7 tests to 2 routing tests that genuinely need a server; remove the pure rendering tests now covered by unit tests
- Add 40 new unit tests across `app-sidebar.test.tsx`, `view-mode-toggle.test.tsx`, `-about.test.tsx`, and `-index.test.tsx`
- Update AGENTS.md to state the unit-test-first preference and reference `docs/testing.md`

Closes: #231

## Test plan
- [x] `make test-ui` passes — 146 unit tests, 0 failures
- [x] `make generate` passes — TypeScript type-checks cleanly

🤖 Generated with [Claude Code](https://claude.com/claude-code) · agent-5

### DIFF
--- a/frontend/e2e/auth.spec.ts
+++ b/frontend/e2e/auth.spec.ts
@@ -20,12 +20,16 @@ import {
  */
 
 test.describe('Authentication', () => {
-  test('should redirect unauthenticated users through OIDC', async ({ page }) => {
+  test('should auto-login unauthenticated users via OIDC', async ({ page }) => {
     await page.goto('/')
 
-    // Root redirects to /profile; auth layout then triggers OIDC auto-redirect
-    // to Dex (PR #230 removed the Sign In button in favour of automatic redirect)
-    await expect(page).toHaveURL(/\/dex\//, { timeout: 15000 })
+    // Root → /profile → OIDC auto-redirect (PR #230 removed the Sign In button).
+    // With --enable-insecure-dex the Dex auto-connector completes auth without a
+    // login form, so the full redirect chain (/ → /dex/auth → /pkce/verify → /profile)
+    // resolves faster than Playwright can poll for /dex/. Verify the end state instead:
+    // the user is authenticated and the profile page content is visible.
+    await expect(page).toHaveURL(/\/profile/, { timeout: 15000 })
+    await expect(page.getByText('ID Token Status')).toBeVisible({ timeout: 10000 })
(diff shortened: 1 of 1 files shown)


jeffmccune:
## Rationale

The core problem is that 4 of 6 E2E spec files depended on live Kubernetes resources, forcing serial execution and 30-second server startup on every CI run.  The fix is straightforward: mock the ConnectRPC query hook layer (`@/queries/*`) with `vi.mock()` + `vi.fn()`, the same pattern already used for `useAuth`, `useOrg`, and `useVersion`.

**Why mock the query hook layer rather than the transport?**  The alternative — providing a mock transport to `TransportProvider` — would require wrapping every test component in a full provider tree and would leak implementation details (ConnectRPC wire format) into tests that should only care about UI behaviour.  Mocking `@/queries/secrets` directly returns controlled data in the exact shape the component expects, with zero infrastructure.

**Why keep `secrets.spec.ts` and most of `auth.spec.ts` as E2E?**  Secret create/read/update/delete against a real K8s API server gives meaningful regression confidence that no mock can provide.  The OIDC login flow (Dex round-trip, PKCE code exchange) similarly cannot be replicated without a real identity provider.

**Why delete `copy-toast.spec.ts` outright?**  The clipboard + toast behavior is already tested in `secret-data-grid.test.tsx` (the "copy button copies the value and shows a toast" test).  The E2E spec was duplicating that coverage while requiring a full cluster.

## Alternatives considered

| Option | Verdict |
|--------|---------|
| Mock at the HTTP transport level (`createConnectTransport` stub) | Rejected — ties tests to wire format, requires provider wrapping everywhere |
| Keep E2E for all navigation tests | Rejected — these tests only check DOM presence, which is inherently unit-testable |
| Migrate `secrets.spec.ts` sharing-panel tests to unit | Partially done — `sharing-panel.test.tsx` already has 30+ cases covering all sharing logic; the E2E tests that remain are the K8s CRUD path |

Issue #245, opened 2026-03-31
feat: fix sidebar Settings links and add org-level nav group


jeffmccune:
Part of #239. Depends on the settings pages created in #239 phases 3 and 4.

## Context

`app-sidebar.tsx` has two problems:

1. The **Settings** nav item under a project links to `/projects/$projectName` (the layout route), which redirects to secrets — so Settings and Secrets go to the same page.
2. There is no org-level navigation. When an org is selected, users can only reach org settings by knowing the URL; there is no sidebar affordance.

## Goal

- Fix the project Settings link to point to the new `/projects/$projectName/settings` route.
- Add an org-level nav group to the sidebar so users can navigate to Org Settings (and, coordinating with #238, to the Projects index).

### 1 — Fix project Settings link

```diff
- { title: "Settings", url: `/projects/${selectedProject}`,          icon: Settings }
+ { title: "Settings", url: `/projects/${selectedProject}/settings`, icon: Settings }
```

### 2 — Add org-level nav group

When `selectedOrg` is set, render a nav group **above** the existing project nav group:

```
── {orgDisplayName} ─────────────────────────
  📁 Projects     /orgs/{orgName}/projects    (coordinates with #238)
  ⚙  Settings     /orgs/{orgName}/settings
```

If `selectedOrg` is null, the org group is hidden entirely.

The group header should display the org's display name (or slug if display name is empty), consistent with how the project picker shows names.

Use the same `SidebarGroup` / `SidebarGroupLabel` / `SidebarMenu` / `SidebarMenuItem` / `SidebarMenuButton` primitives already used for the project nav items.

### 3 — Active state

`SidebarMenuButton` accepts an `isActive` prop. Mark items active when `useRouterState().location.pathname` starts with the item's URL prefix, consistent with the pattern already used in the sidebar.

### RED — update sidebar tests

In `frontend/src/components/app-sidebar.test.tsx`, add assertions:

- When `selectedProject` is set, the Settings link href is `/projects/$projectName/settings`
- When `selectedOrg` is set (no project), the org Settings link `/orgs/$orgName/settings` is present
- When `selectedOrg` is set (no project), the Projects link `/orgs/$orgName/projects` is present
- When `selectedOrg` is null, no org-level group is rendered

### GREEN — update the component

Implement the changes in `app-sidebar.tsx` to make the tests pass.

## Files to Modify

| File | Change |
|------|--------|
[...]

PR #251, opened 2026-03-31
feat: fix sidebar Settings links and add org-level nav group

jeffmccune:
## Summary
- Add `SidebarGroupLabel` showing org display name above org nav items
- Render org nav group independently so both org and project groups are visible when both are selected
- Fix tests to use resolved hrefs and cover new org group label behavior

Closes: #245

## Test plan
- [x] All 261 UI unit tests pass (`make test-ui`)
- [x] `make generate` succeeds with no TypeScript errors
- [x] Visual verification screenshots captured and committed

🤖 Generated with [Claude Code](https://claude.com/claude-code) · agent-0


jeffmccune:
## Rationale

The issue identified two gaps in sidebar navigation:

1. **Project Settings link was broken** — it pointed to `/projects/$projectName` (the layout route), which redirected to Secrets. Fixed by pointing to `/projects/$projectName/settings/`.

2. **No org-level nav when project selected** — the previous implementation used an if-else that showed either org items OR project items, never both. This meant when a user had a project selected they could not reach org Settings without knowing the URL.

## Implementation approach

Replaced the single `navItems` array with two separate arrays — `orgNavItems` and `projectNavItems` — each rendered in their own `SidebarGroup`. The org group always renders when an org is selected (regardless of project selection) and includes a `SidebarGroupLabel` with the org's display name. The project group renders only when a project is selected.

## Alternatives considered

| Option | Verdict |
|--------|---------|
| Keep if-else, only show org group when no project selected | Rejected — users can't navigate to org Settings while working in a project |
| Add a back-to-org breadcrumb instead of a persistent nav group | Rejected — the issue explicitly requests a sidebar nav group |

## Visual verification

Org nav group with label and Settings active

Screenshot shows: org "acme" group label, Projects and Settings links, Settings highlighted as active on the `/orgs/acme/settings` page.

Issue #355, opened 2026-04-05
feat(proto): add creator_email and created_at fields to Organization and Project messages


jeffmccune:
## Parent Issue

Part of #354

## Goal

Add `creator_email` and `created_at` fields to the `Organization` and `Project` proto messages so that the backend can populate them and the frontend can display them. This is the proto-first foundation that all subsequent phases depend on.

## Acceptance Criteria

- [ ] `Organization` message has `string creator_email` and `google.protobuf.Timestamp created_at` fields (or `string created_at` in RFC3339 format — see implementation notes).
- [ ] `Project` message has `string creator_email` and `google.protobuf.Timestamp created_at` fields (or `string created_at`).
- [ ] Field numbers do not conflict with existing fields and existing fields are untouched.
- [ ] `make generate` succeeds with no errors (regenerates Go and TypeScript types).
- [ ] `make test` passes.

### Field numbering

Current last fields:
- `Organization`: field 8 (`repeated ShareGrant default_role_grants = 8`) → add at 9, 10
- `Project`: field 9 (`repeated ShareGrant default_role_grants = 9`) → add at 10, 11

### Timestamp representation

Prefer `string created_at` (RFC3339) over `google.protobuf.Timestamp` to keep the proto simple and avoid importing the well-known types package unless it is already imported. Check existing imports in both proto files before deciding. Either approach is acceptable; document the choice in the PR.

### Files to modify

- `proto/holos/console/v1/organizations.proto` — add `creator_email` and `created_at` to `Organization` message
- `proto/holos/console/v1/projects.proto` — add `creator_email` and `created_at` to `Project` message

After editing, run:
```bash
make generate
```

Confirm that `gen/` and `frontend/src/gen/` are updated and that `make test` passes.

### Testing approach

No new tests in this phase — generated code is verified by `make generate` succeeding and `make test` passing. Backend and frontend tests come in later phases.

### Dependencies

None. This is the first phase.

Issue #354, opened 2026-04-05
feat: store and display resource creator and creation time


jeffmccune:
## Problem

When an organization or project is created, there is no record of who created it or when. Operators need this information for auditing and governance. The creator's email (from the OIDC ID token) and the creation timestamp should be persisted as annotations on the Kubernetes Namespace and surfaced as read-only fields in the frontend settings view.

## Acceptance Criteria

- [ ] When an organization is created, the creator's email (from the ID token) is stored as a `console.holos.run/creator-email` annotation on the Kubernetes Namespace.
- [ ] When a project is created, the creator's email is stored the same way.
- [ ] The `Organization` and `Project` proto messages include `creator_email` and `created_at` fields.
- [ ] `GetOrganization` and `GetProject` RPCs populate these fields from namespace annotations/metadata.
- [ ] `ListOrganizations` and `ListProjects` also populate these fields.
- [ ] The organization settings page shows read-only "Creator" and "Created" fields.
- [ ] The project settings page shows read-only "Creator" and "Created" fields.
- [ ] These fields are not editable (no update RPC changes required).
- [ ] Existing namespaces without these annotations degrade gracefully (empty/unknown display).

## Implementation Plan

This issue tracks the full implementation. Sub-issues cover each phase:

- [ ] #355 — feat(proto): add creator_email and created_at fields to Organization and Project messages
- [ ] #356 — feat(backend): store creator email annotation and populate proto fields on read
- [ ] #357 — feat(frontend): show read-only creator and creation time in org and project settings
- [ ] #358 — chore: cleanup stale code and docs after creator/creation-time feature

Implement phases in order. Each phase should leave the codebase in a working state.

PR #359, opened 2026-04-05
feat(proto): add creator_email and created_at fields to Organization and Project messages

jeffmccune:
## Summary
- Add `creator_email` (string) and `created_at` (string, RFC3339) fields to `Organization` proto message at field numbers 9 and 10
- Add `creator_email` (string) and `created_at` (string, RFC3339) fields to `Project` proto message at field numbers 10 and 11
- Regenerate Go and TypeScript types via `make generate`
- All existing tests continue to pass

Closes: #355

## Test plan
- [x] `make generate` succeeds with no errors
- [x] `make test` passes (all 396 UI tests + all Go tests)

## Implementation note
`string` fields were used instead of `google.protobuf.Timestamp` because neither `organizations.proto` nor `projects.proto` currently imports the well-known types package. Using plain RFC3339 strings avoids a new import dependency and keeps the proto files consistent with their current style.


jeffmccune:
## Rationale

The issue asked for `creator_email` and `created_at` on both `Organization` and `Project` as the proto-first foundation for displaying creator/timestamp info in the UI (parent issue #354). This PR is intentionally narrow: it only touches the proto definitions and the regenerated Go/TypeScript artifacts.

### Why `string` instead of `google.protobuf.Timestamp`?

The issue explicitly preferred `string created_at` in RFC3339 format to keep the proto simple. Neither `organizations.proto` nor `projects.proto` currently imports `google/protobuf/timestamp.proto`, so adding `google.protobuf.Timestamp` would introduce a new import and pull in the well-known types package. Since RFC3339 strings work fine for display purposes and are already used in many Go/K8s APIs, the string approach was chosen.

### Field number placement

Field numbers follow the convention of appending after the last existing field:
- `Organization`: fields 9 (`creator_email`) and 10 (`created_at`), after the existing field 8 (`default_role_grants`)
- `Project`: fields 10 (`creator_email`) and 11 (`created_at`), after the existing field 9 (`default_role_grants`)

No existing fields were modified, so there is no wire-format breakage.

## Alternatives considered

| Option | Verdict |
|--------|---------|
| `google.protobuf.Timestamp` for `created_at` | Rejected — would introduce a new import, and RFC3339 strings suffice for display |
| Add a `creator_display_name` field too | Out of scope for this phase; the issue only asks for `creator_email` |
| Add fields to `CreateOrganizationRequest`/`CreateProjectRequest` | Not needed — `creator_email` is populated by the backend from the authenticated user's JWT, not passed by the client |

Issue #396, opened 2026-04-05
refactor(frontend): convert Create Deployment modal to a dedicated page


jeffmccune:
## Parent Issue

Part of #394

## Goal

Replace the Create Deployment modal (currently inline in the deployments index page) with a dedicated page at `/projects/$projectName/deployments/new`. The modal has too many form fields (8+) for a dialog, and users lose their work by accidentally clicking outside.

## Acceptance Criteria

- [ ] New route at `/_authenticated/projects/$projectName/deployments/new` renders a Create Deployment page
- [ ] Page contains all existing form fields: Display Name, Name (slug), Description, Template select, Image, Tag, Command (StringListInput), Args (StringListInput), Environment Variables (EnvVarEditor)
- [ ] "Create Deployment" button on deployments list page navigates to the new page (not a modal)
- [ ] Empty state on deployments list page also links to the new page
- [ ] When no templates exist, show a link to `/projects/$projectName/templates/new` instead of opening a sub-modal
- [ ] On successful creation, navigates to the deployment detail page (`/projects/$projectName/deployments/$deploymentName`)
- [ ] Form validation preserved (name, template, image, tag required)
- [ ] Unit tests for the new page component with mocked query hooks
- [ ] Tests pass: `make test`

### Files to create

- `frontend/src/routes/_authenticated/projects/$projectName/deployments/new.tsx` — the new Create Deployment page route
- `frontend/src/routes/_authenticated/projects/$projectName/deployments/-new.test.tsx` — unit tests

### Files to modify

- `frontend/src/routes/_authenticated/projects/$projectName/deployments/index.tsx` — remove inline modal code (~120 lines), replace "Create Deployment" button with `<Link to="new">` navigation

### Key patterns to follow

- Use `createFileRoute('/_authenticated/projects/$projectName/deployments/new')` for the route definition
- Use `useNavigate()` to redirect after successful creation
- Use `useCreateDeployment` from `@/queries/deployments` and `useListDeploymentTemplates` from `@/queries/deployment-templates`
- Reuse existing components: `StringListInput`, `EnvVarEditor`, `filterEnvVars`
- When no templates exist, render a link: "No templates available. [Create a template](/projects/$projectName/templates/new) first."
- The page layout gives much more room for the env var editor and command/args fields

### Testing approach

- Mock `@/queries/deployments` and `@/queries/deployment-templates` with `vi.mock()` and `vi.fn()`
- Test form rendering, validation (required fields), submission with all fields, template selection
[...]

PR #421, opened 2026-04-05
refactor(frontend): convert Create Template modal to a dedicated page

jeffmccune:
## Summary
- Add dedicated Create Template page at `/_authenticated/projects/$projectName/templates/new`
- Page contains all form fields: Display Name, Name slug with auto-derivation, Description, CUE Template textarea, Preview toggle with rendered JSON
- Replace modal trigger buttons on templates list with `<Link>` navigation to the new page
- On successful creation, navigate to the template detail page
- Form validation preserved (name required, error display)
- 15 new unit tests for the Create Template page component
- Existing index tests updated to reflect navigation-based behavior instead of modal dialog

Closes: #395

## Test plan
- [ ] `make test-ui` passes (451 tests across 35 files)
- [ ] `make generate` builds without TypeScript errors
- [ ] Create Template button on templates list navigates to `/templates/new`
- [ ] Empty state link also navigates to `/templates/new`
- [ ] CUE editor has more vertical space (not constrained by a modal)
- [ ] Cancel link returns to templates list

🤖 Generated with [Claude Code](https://claude.com/claude-code) · agent-aae824f2

PR #427, opened 2026-04-05
refactor(frontend): convert Create Deployment modal to dedicated page

jeffmccune:
## Summary
- Add new route at `/_authenticated/projects/$projectName/deployments/new` with a full-page Create Deployment form
- Replace inline Create Deployment modal in the deployments index page with Link navigation to the new page
- Remove sub-modal pattern for creating templates; when no templates exist, show a link to `/projects/$projectName/templates/new` instead
- Remove now-dead `CreateTemplateModal` component and its tests (13 fewer tests)
- Add 25 unit tests for the new Create Deployment page
- Update 17 index tests to reflect the navigation-based pattern

Closes: #396

## Test plan
- [x] `make test` passes (446 tests, 35 files)
- [x] `make generate` passes (TypeScript build succeeds)
- [x] Create Deployment button/link navigates to new page
- [x] Empty state also links to new page
- [x] No templates state shows link to templates/new
- [x] Successful creation navigates to deployment detail page
- [x] Form validation works for all required fields

🤖 Generated with [Claude Code](https://claude.com/claude-code) · agent-aca14464


jeffmccune:
## Rationale

The Create Deployment modal had 8+ form fields (Display Name, Name slug, Description, Template select, Image, Tag, Command, Args, Environment Variables) — far too many for a dialog. Users could accidentally dismiss it by clicking outside, losing all their work. The nested CreateTemplateModal sub-modal added extra complexity.

## Approach

The same pattern used in PR #421 (Create Template modal → page) was applied here:
- New route `deployments/new.tsx` with the same Card/CardContent layout pattern
- `index.tsx` simplified to a list-only page with navigation links
- The `CreateTemplateModal` component became dead code and was removed

## Navigation after creation

The `CreateDeploymentResponse` proto has a top-level `name` field (not nested), so the navigation uses `result.name` to get the created deployment name for the redirect to `/projects/$projectName/deployments/$deploymentName`.

## No-templates state

Instead of opening a sub-modal (which required `CreateTemplateModal`), the new page shows an inline message with a direct link to `/projects/$projectName/templates/new`. This is simpler and avoids modal nesting entirely.

## Alternatives considered

| Option | Verdict |
|--------|---------|
| Keep modal but make it scrollable | Rejected — modals are not the right pattern for complex forms with many fields |
| Keep `CreateTemplateModal` but just not use it | Rejected — dead code should be removed |
| Require K8s fixtures for E2E test | Rejected — behavior is fully verifiable with unit tests |

Issue #423, opened 2026-04-05
feat(proto): add port field to deployment and template proto messages


jeffmccune:
## Parent Issue

Part of #422

## Goal

Add `int32 port` to the deployment proto messages so the container port is part of the API contract. This is the foundational change that all subsequent phases depend on.

## Acceptance Criteria

- [ ] `Deployment` message has `int32 port = 13`
- [ ] `CreateDeploymentRequest` message has `int32 port = 11`
- [ ] `UpdateDeploymentRequest` message has `optional int32 port = 10`
- [ ] `DeploymentDefaults` message has `int32 port = 6`
- [ ] `make generate` succeeds (Go and TypeScript types regenerated)
- [ ] `make test` passes (existing tests still compile and pass)

### Files to modify

- `proto/holos/console/v1/deployments.proto` — add `port` to `Deployment`, `CreateDeploymentRequest`, `UpdateDeploymentRequest`
- `proto/holos/console/v1/deployment_templates.proto` — add `port` to `DeploymentDefaults`

### Field number assignment

Use the next available field number in each message:
- `Deployment`: field 13 (after `env = 12`)
- `CreateDeploymentRequest`: field 11 (after `env = 10`)
- `UpdateDeploymentRequest`: field 10 (after `env = 9`)
- `DeploymentDefaults`: field 6 (after `env = 5`)

### Testing approach

No new tests needed — this phase only adds proto fields. Run `make generate` then `make test` to verify nothing breaks.

### Dependencies

None — this is the first phase.

Issue #422, opened 2026-04-05
feat(deployments): add container port configuration to deployment form


jeffmccune:
## Problem

The Create Deployment form (and Update Deployment) does not expose any way for the user to specify the port number their container listens on. The CUE default template currently hardcodes `containerPort: 8443` with HTTPS semantics (`name: "https"`, Service port 443). Most user workloads serve plain HTTP on a port like 8080, and users need a way to configure this without editing the CUE template directly.

## Acceptance Criteria

- [ ] `Deployment`, `CreateDeploymentRequest`, `UpdateDeploymentRequest` proto messages include an `int32 port` field
- [ ] `DeploymentDefaults` proto message includes an `int32 port` field so templates can provide a default port
- [ ] Backend `DeploymentInput` struct passes `port` to the CUE renderer
- [ ] Default CUE template `#Input` accepts `port` with a default of `8080`, uses it for `containerPort` and names the port `"http"`; Service exposes port `80` targeting the named port
- [ ] Backend stores and retrieves the port value in the deployment ConfigMap
- [ ] Create Deployment form includes a Port number input, pre-populated from template defaults (or 8080)
- [ ] Update Deployment form includes the Port number input
- [ ] Go unit tests cover port round-trip (store, retrieve, render)
- [ ] Frontend unit tests cover port input rendering and validation
- [ ] `make generate` and `make test` pass

## Implementation Plan

This issue tracks the full implementation. Sub-issues cover each phase:

- [ ] #423 — Phase 1: Proto definitions — add `port` field to deployment and template messages
- [ ] #424 — Phase 2: Backend — wire port through handler, K8s storage, CUE renderer, and default template
- [ ] #425 — Phase 3: Frontend — add port input to Create and Update Deployment forms
- [ ] #426 — Phase 4: Cleanup — remove stale hardcoded port references and update docs

Implement phases in order. Each phase should leave the codebase in a working state.

PR #434, opened 2026-04-05
feat(proto): add port field to deployment and template proto messages

jeffmccune:
## Summary
- Add `int32 port = 13` to `Deployment` message
- Add `int32 port = 11` to `CreateDeploymentRequest` message
- Add `optional int32 port = 10` to `UpdateDeploymentRequest` message
- Add `int32 port = 6` to `DeploymentDefaults` message
- Regenerate Go and TypeScript protobuf types

Closes: #423

## Test plan
- [x] `make generate` succeeds
- [x] `make test` passes (all 451 UI tests + all Go tests pass)


jeffmccune:
## Rationale

This PR adds `port` as a first-class field to the deployment and template proto messages as the foundational change for the container port feature (#422).

The field is added as a plain `int32` (not `optional`) on `Deployment` and `CreateDeploymentRequest` since a zero value (0) is a safe default meaning "not set", matching the proto3 convention. On `UpdateDeploymentRequest` it is `optional int32` so the backend can distinguish "caller didn't touch the port" from "caller explicitly set port to 0 to clear it".

In `DeploymentDefaults`, `port` is also a plain `int32` since defaults are sparse — a zero means no default is provided.

## Alternatives considered

| Option | Verdict |
|--------|---------|
| Use `optional int32` everywhere | Rejected — `optional` adds HasPresence overhead; for `Deployment` and `CreateDeploymentRequest`, zero is an unambiguous "not configured" sentinel |
| Add a `repeated int32 ports` list | Rejected — the issue specifies a single `port` field; multi-port support is out of scope for this phase |
| Name it `container_port` | Rejected — the issue specifies `port`; shorter names are idiomatic in this codebase |

Issue #447, opened 2026-04-05
feat(proto): add cue_system_input field to RenderDeploymentTemplateRequest


jeffmccune:
## Parent Issue

Part of #445

## Goal

Add a `cue_system_input` field to `RenderDeploymentTemplateRequest` in the proto definition, and split the Go `DeploymentInput` struct into `SystemInput` (with Claims) and `UserInput`. Run `make generate` to produce updated Go and TypeScript types.

## Acceptance Criteria

- [ ] `RenderDeploymentTemplateRequest` has a new `cue_system_input` string field for system-provided CUE values
- [ ] Go structs in `console/deployments/render.go` split into:
  - `ClaimsInput` — OIDC claims: `Iss`, `Sub`, `Aud`, `Exp`, `Iat`, `Email`, `EmailVerified`, `Name`, `Groups`
  - `SystemInput` — `Project`, `Namespace`, `Claims ClaimsInput`
  - `UserInput` — `Name`, `Image`, `Tag`, `Command`, `Args`, `Env`, `Port`
- [ ] `make generate` succeeds
- [ ] `make build` succeeds (compilation, no runtime changes yet)

### Proto change

In `proto/holos/console/v1/deployment_templates.proto`, add to `RenderDeploymentTemplateRequest`:

```protobuf
message RenderDeploymentTemplateRequest {
  string cue_template = 2;
  // cue_system_input contains CUE source for system-provided values (project, namespace, claims).
  // These values are trusted — set by the console backend from authenticated context.
  // The claims field carries the OIDC ID token claims of the authenticated user.
  string cue_system_input = 7;
  // cue_input contains CUE source for user-provided deployment parameters
  // (name, image, tag, command, args, env, port).
  string cue_input = 6;
}
```

Use field number 7 for `cue_system_input` (next available).

### Go struct split

In `console/deployments/render.go`:

```go
// ClaimsInput carries OIDC ID token claims passed to CUE templates.
type ClaimsInput struct {
    Iss           string   `json:"iss"`
    Sub           string   `json:"sub"`
    Aud           string   `json:"aud,omitempty"`
    Exp           int64    `json:"exp"`
    Iat           int64    `json:"iat"`
    Email         string   `json:"email"`
    EmailVerified bool     `json:"email_verified"`
    Name          string   `json:"name,omitempty"`
    Groups        []string `json:"groups,omitempty"`
}

// SystemInput contains trusted values set by the console backend.
type SystemInput struct {
[...]

PR #454, opened 2026-04-05
feat(proto): add cue_system_input field to RenderDeploymentTemplateRequest

jeffmccune:
## Summary

- Add `cue_system_input` (field 7) to `RenderDeploymentTemplateRequest` proto for system-provided CUE values
- Split `DeploymentInput` into three structs in `console/deployments/render.go`:
  - `ClaimsInput` — OIDC token claims (iss, sub, aud, exp, iat, email, email_verified, name, groups)
  - `SystemInput` — trusted backend values (project, namespace, claims)
  - `UserInput` — user-provided deployment params (name, image, tag, command, args, env, port)
- Update `Render()` signature from `DeploymentInput` to `(SystemInput, UserInput)`
- Update `evaluate()` to fill `"input"` path from `UserInput` and `"system"` path from `SystemInput`
- Update `evaluateWithCueInput()` to read `system.namespace` instead of `input.namespace`
- Update `default_template.cue` to use `system.namespace` and define `#System`/`#Claims` constraints
- Update all tests to use the new split input types
- Update `docs/cue-template-guide.md` to reflect new schema and remove stale "Planned Extensions" section

Closes: #447

## Test plan
- [x] `make generate` succeeds (proto + TypeScript types regenerated)
- [x] `make build` succeeds (compilation)
- [x] All Go tests pass (`go test ./...`)
- [x] No frontend changes (no UI screenshots required)


jeffmccune:
## Rationale

This PR implements the structural foundation for the system/user input trust boundary described in ADR 013 (merged in PR #453).

### Why split DeploymentInput into SystemInput + UserInput?

The core issue: the old `DeploymentInput` struct mixed user-controlled fields (`name`, `image`, `tag`) with server-controlled fields (`project`, `namespace`). Both were serialized into the same CUE `input` path, meaning CUE templates could not distinguish trusted from user-supplied values.

After this split:
- `system.namespace` is guaranteed to come from `Resolver.ProjectNamespace()` — no user can override it
- `system.claims` carries verified OIDC token claims — not user-supplied data
- `input.name`, `input.image`, etc. are explicitly user-controlled

### Why update default_template.cue now?

The default template references `input.namespace` extensively. If we kept the old template but changed evaluate() to put namespace in `system`, all existing tests using the default template would fail. Updating the template and tests in the same PR keeps everything consistent and verifiable.

### Why update evaluateWithCueInput() too?

The adapter (used for the `RenderDeploymentTemplate` preview RPC) uses `RenderWithCueInput()` which reads the namespace from the unified CUE value. After the split, the namespace lives at `system.namespace` not `input.namespace`. Updated accordingly to avoid breaking the template preview flow.

## Alternatives considered

| Option | Verdict |
|--------|---------|
| Keep `DeploymentInput` and add parallel `SystemInput` | Rejected — would require callers to pass both types and maintain two overlapping structs |
| Make `evaluate()` accept a single combined struct | Rejected — defeats the trust boundary purpose; system fields could be set by user |
| Keep `input.namespace` as an alias alongside `system.namespace` | Rejected — creates confusion about which is authoritative |

Issue #448, opened 2026-04-05
feat(templates): update Go rendering pipeline and CUE template for split system/user input


jeffmccune:
## Parent Issue

Part of #445

## Goal

Update the Go rendering pipeline to accept separate system and user inputs (including OIDC claims), update the default CUE template to use `system` and `input` as distinct fields with a `#Claims` schema, and update all RPC handlers to populate claims from the authenticated context.

## Acceptance Criteria

- [ ] `CueRenderer.Render()` accepts `SystemInput` and `UserInput` separately, filling `system` and `input` CUE paths
- [ ] `CueRenderer.RenderWithCueInput()` accepts both `cueSystemInput` and `cueInput` strings and unifies them at `system` and `input` paths respectively
- [ ] Default CUE template defines `#SystemInput` with `project`, `namespace`, and `claims: #Claims`
- [ ] Default CUE template defines `#Claims` validating standard OIDC claims: `iss` (string), `sub` (string), `iat` (int), `exp` (int), `email` (string), `email_verified` (bool), optional `name`, optional `groups`, open struct for provider-specific claims
- [ ] Default CUE template uses `system.project`, `system.namespace` instead of `input.project`, `input.namespace`
- [ ] Default CUE template adds `console.holos.run/deployer-email: system.claims.email` annotation to all resources via `_labels` (or a separate `_annotations` helper)
- [ ] `RenderDeploymentTemplate` RPC handler passes `cue_system_input` to the renderer
- [ ] `CreateDeployment` handler builds `SystemInput` (including `ClaimsInput` from `rpc.ClaimsFromContext(ctx)` and `IDToken` standard fields) and `UserInput` separately
- [ ] Go unit tests verify rendering with split inputs including claims
- [ ] Go unit tests verify that `system.claims.email` appears in rendered resource annotations
- [ ] `make test-go` passes
- [ ] `make generate` passes (UI build may fail until frontend is updated — that's OK for this phase)

### CUE template changes

Update `console/templates/default_template.cue`:

```cue
// #Claims defines the OIDC ID token claims provided by the console backend.
// Standard claims are validated; additional provider-specific claims are allowed.
#Claims: {
    iss:            string
    sub:            string
    iat:            int
    exp:            int
    email:          string
    email_verified: bool
    name?:          string
    groups?: [...string]
    ...  // allow provider-specific claims
}

#SystemInput: {
    project:   string
    namespace: string
    claims:    #Claims
}

#Input: {
[...]

Issue #445, opened 2026-04-05
feat(templates): separate system and user input in CUE template interface


jeffmccune:
## Problem

The CUE template interface currently uses a single `input` field that mixes system-provided values (`project`, `namespace`) with user-provided values (`name`, `image`, `tag`, `port`, `command`, `args`, `env`). This makes it difficult for template authors to distinguish trusted input (set by the console backend from authenticated context) from less-trusted input (provided by the user via the deployment form).

Separating these into distinct fields (`system` and `input`) establishes a clear trust boundary:
- **`system`**: Set by the console backend. Template authors can trust these values for security-sensitive decisions (namespace constraints, RBAC-derived policy).
- **`input`**: User-provided deployment parameters. Template authors should validate/constrain these values.

Additionally, the frontend template editor's preview tab currently shows a single CUE input textarea that mixes both categories. Splitting it into two clearly labeled areas reinforces the trust model for template authors testing their templates.

## Acceptance Criteria

- [ ] ADR documents the trust-boundary rationale for separating system and user input
- [ ] CUE template `#Input` schema split into `#SystemInput` (project, namespace) and `#Input` (name, image, tag, port, command, args, env)
- [ ] Default template uses `system.project`, `system.namespace` instead of `input.project`, `input.namespace`
- [ ] `RenderDeploymentTemplateRequest` proto has separate `cue_system_input` and `cue_input` fields
- [ ] Go rendering pipeline (`DeploymentInput` struct, `Render`, `RenderWithCueInput`) splits system and user input
- [ ] Frontend template editor preview tab shows two distinct input areas: "System Input" (read-only or clearly labeled as system-provided) and "User Input" (editable)
- [ ] Frontend new-template page preview also uses the split input pattern
- [ ] `AGENTS.md` guardrails updated: new fields added to template-related RPCs must also be reflected in the CUE template editor preview UI
- [ ] `docs/cue-template-guide.md` updated to document the split interface
- [ ] All existing tests pass; new tests cover the split interface
- [ ] Cleanup phase removes stale references to the old single-input pattern

## Implementation Plan

This issue tracks the full implementation. Sub-issues cover each phase:

- [ ] #446 — ADR for separating system and user input trust boundary
- [ ] #447 — Proto changes: add `cue_system_input` field, split Go structs
- [ ] #448 — Backend: update Go rendering pipeline and default CUE template
- [ ] #449 — Frontend: split template editor preview into system and user input areas
[...]

PR #455, opened 2026-04-05
feat(templates): update Go rendering pipeline and CUE template for split system/user input

jeffmccune:
## Summary
- Add `Iss`, `Iat`, `Exp` fields to `rpc.Claims` and populate from `idToken.Issuer`, `idToken.IssuedAt`, `idToken.Expiry` in `extractAndVerifyToken`
- Update default CUE template: make `#Claims` fields required (iss, sub, exp, iat, email, email_verified) with open struct for provider-specific claims; add `_annotations` helper with `console.holos.run/deployer-email` from `system.claims.email`; apply annotations to all resource metadata
- Update `Renderer` interface in `console/templates/handler.go` to accept separate `cueSystemInput` and `cueInput` parameters
- Update `CueRendererAdapter.Render` to combine the separate inputs into a single CUE document
- Update `RenderDeploymentTemplate` handler to forward `req.Msg.CueSystemInput`
- Update `CreateDeployment` and `UpdateDeployment` to populate `Iss`, `Iat`, `Exp` in `ClaimsInput`
- Add unit tests: `TestCueRenderer_ClaimsPropagation`, `TestDefaultTemplate_DeployerEmailAnnotation`, `TestDefaultTemplate_ClaimsEmailInAnnotation`

Closes: #448

## Test plan
- [x] `make test-go` passes
- [x] `make generate` passes
- [x] `TestCueRenderer_ClaimsPropagation` verifies system.claims.email appears in rendered resource annotations
- [x] `TestDefaultTemplate_DeployerEmailAnnotation` verifies deployer-email annotation is set on all default template resources
- [x] `TestDefaultTemplate_ClaimsEmailInAnnotation` verifies deployer-email annotation reflects claims.email for different email values


jeffmccune:
## Rationale

This PR implements the Go rendering pipeline changes for the system/user input split (sub-issue 3 of 6 in #445).

### Design decisions

**Claims propagation via `rpc.Claims` extension (Option A)**

The issue offered two options for getting `iss`, `iat`, `exp` into `ClaimsInput`:
- Option A: Extend `rpc.Claims` with `Iss`, `Iat`, `Exp` fields populated from `idToken`
- Option B: Store the full `IDToken` in context alongside claims

Option A was chosen (as recommended) because it keeps the context API simple and type-safe. Adding three fields to `Claims` is minimal and all downstream consumers just get the enriched struct automatically.

**CUE `#Claims` with required fields and open struct**

The default template now enforces standard OIDC claims as required fields (`iss`, `sub`, `exp`, `iat`, `email`, `email_verified`). The `...` open struct suffix allows provider-specific claims through without CUE errors. This follows the principle of validating what we know and not breaking when providers add additional claims.

**CueRendererAdapter: combine then pass to RenderWithCueInput**

Rather than adding a new code path in `CueRenderer`, the adapter concatenates `cueSystemInput` and `cueInput` into a single CUE document. This avoids duplication since `RenderWithCueInput` already handles the top-level unification pattern used by the template preview RPC.

**`_annotations` helper**

Following the same pattern as the existing `_labels` helper, `_annotations` is a top-level CUE definition (prefixed with `_` so it's not exported) that all resources reference. Adding the deployer-email annotation only requires updating one place.

## Alternatives considered

| Option | Verdict |
|--------|---------|
| Add `email_verified` to `#Claims` as optional (`email_verified?: bool`) | Rejected — it's a standard OIDC claim and the backend always sets it from the verified token |
| Add `RenderWithSplitInput(ctx, src, systemSrc, userSrc string)` to CueRenderer | Rejected — the adapter's concatenation approach is simpler and keeps CueRenderer focused |
| Store per-resource annotation logic in Go validation | Rejected — CUE is the right layer for template-defined annotations; Go enforces structural invariants only |

Issue #468, opened 2026-04-06
feat(proto): add SystemTemplateService and PERMISSION_SYSTEM_DEPLOYMENTS_EDIT to rbac.proto


jeffmccune:
## Parent Issue

Part of #467

## Goal

Define the proto interface for system templates and the new org-scoped permission. Running \`make generate\` regenerates Go and TypeScript types so subsequent phases can implement against concrete types.

## Acceptance Criteria

- [ ] \`proto/holos/console/v1/system_templates.proto\` is created with:
  - \`SystemTemplate\` message: \`name\`, \`org\`, \`display_name\`, \`description\`, \`cue_template\`, \`mandatory\` (bool), \`gateway_namespace\` (string, default \`istio-ingress\` via documentation comment)
  - \`SystemTemplateService\` with \`List/Get/Create/Update/Delete\` and \`RenderSystemTemplate\` RPCs mirroring the shape of \`DeploymentTemplateService\`
  - List/Get/Create/Update/Delete request/response messages scoped by \`org\` instead of \`project\`
  - \`RenderSystemTemplateRequest\` / \`RenderSystemTemplateResponse\` matching the existing \`RenderDeploymentTemplateRequest\` shape (same \`cue_template\`, \`cue_system_input\`, \`cue_input\` fields)
- [ ] \`proto/holos/console/v1/rbac.proto\` gains \`PERMISSION_SYSTEM_DEPLOYMENTS_EDIT = 32\` (next available value, does not renumber existing values)
- [ ] \`make generate\` succeeds — generated Go types in \`gen/\` and TypeScript types in \`frontend/src/gen/\` are committed
- [ ] No implementation yet — only proto + generated code

### Files to create/modify

- \`proto/holos/console/v1/system_templates.proto\` — new file; import \`deployments.proto\` for \`EnvVar\` reuse and \`rbac.proto\`
- \`proto/holos/console/v1/rbac.proto\` — append \`PERMISSION_SYSTEM_DEPLOYMENTS_EDIT = 32\` to the \`Permission\` enum
- Run \`make generate\` and commit the results under \`gen/\` and \`frontend/src/gen/\`

### Key design decisions

- \`SystemTemplate.org\` scopes the template to an organization namespace (not a project namespace)
- \`mandatory: bool\` — when true, template is auto-applied at project creation; cannot be modified by users without \`PERMISSION_SYSTEM_DEPLOYMENTS_EDIT\`
[...]

PR #476, opened 2026-04-06
feat(proto): add SystemTemplateService and PERMISSION_SYSTEM_DEPLOYMENTS_EDIT to rbac.proto

jeffmccune:
## Summary
- Add `proto/holos/console/v1/system_templates.proto` defining `SystemTemplateService` with List/Get/Create/Update/Delete/Render RPCs and `SystemTemplate` message with `mandatory` bool and `gateway_namespace` string fields
- Add `PERMISSION_SYSTEM_DEPLOYMENTS_EDIT = 32` to the `Permission` enum in `rbac.proto` for org-level OWNER-only access control
- Regenerate Go types (`gen/`) and TypeScript types (`frontend/src/gen/`) via `make generate`

Closes: #468

## Test plan
- [x] `make generate` succeeds with no errors
- [x] `make test` passes (all 36 UI test files, all 12 Go packages)
- [ ] Subsequent phases implement backend handler, RBAC, and frontend


jeffmccune:
## Rationale

This PR follows the proto-first pattern established in the codebase: define the interface in proto, regenerate, then implement backend and frontend in subsequent phases. This ensures all generated types are available before any implementation work begins.

## Design decisions

**`SystemTemplate.org` instead of `project`**: System templates are org-scoped resources, stored in the org namespace ConfigMap (not per-project). The `org` field mirrors the `project` field in `DeploymentTemplate`.

**`mandatory: bool`**: A simple flag indicating the template should be auto-applied at project creation time. No complex scheduling or ordering — just a boolean gate.

**`gateway_namespace: string`**: Carries the target Gateway namespace for cross-namespace routing scenarios (e.g., Istio ReferenceGrant). Defaults to `istio-ingress` per Istio's recommended Helm install convention, but is configurable per template.

**`PERMISSION_SYSTEM_DEPLOYMENTS_EDIT = 32`**: Follows the sparse permission numbering pattern. Value 32 is the next sequential value after `PERMISSION_PROJECT_DEPLOYMENTS_ENABLE = 31`. This permission is intentionally scoped to org-level OWNER cascade only (not inherited by project-level OWNERs) — enforced in phase 2.

**`RenderSystemTemplate` reuses the same request shape as `RenderDeploymentTemplate`**: The `cue_template`, `cue_system_input`, and `cue_input` fields are identical to simplify backend implementation and frontend reuse.

## Alternatives considered

| Option | Verdict |
|--------|---------|
| Reuse `DeploymentTemplateService` with an `is_system` flag | Rejected — system templates have different RBAC, different storage location (org ns vs project ns), and different lifecycle. Separate service is cleaner. |
| Add `gateway_namespace` to `DeploymentDefaults` | Rejected — this field is system-template-specific, not a deployment default. Belongs on `SystemTemplate` directly. |
| Use `optional bool mandatory` | Rejected — `mandatory: false` is always a valid explicit value. No need for optionality here. |

PR #653, opened 2026-04-09
feat(folders): implement FolderService with hierarchy enforcement

jeffmccune:
## Summary
- Create `console/folders/` package with full FolderService implementation
- Enforce 3-folder depth limit at `CreateFolder` time (rejects depth ≥ 4 with `InvalidArgument`)
- Default-share cascade: merges ancestor `default-share-users`/`default-share-roles` into new folder's initial grants
- `DeleteFolder` guards against child folders and child projects (`FailedPrecondition`)
- Wire FolderService into `console/console.go`
- Full unit test coverage for all RPCs and depth scenarios

Closes: #628

## Test plan
- [x] `go test -count=1 ./console/folders/...` passes (k8s + handler tests)
- [x] `go test -count=1 ./console/... ./api/...` passes (all 15 packages)
- [x] Tests cover: depth 1/2/3 creation, depth 4 rejection, deletion with children, default-share cascade, creator auto-owner, sharing updates, parent type resolution

### DIFF
--- a/console/console.go
+++ b/console/console.go
@@ -33,6 +33,7 @@ import (
 	"golang.org/x/net/http2/h2c"
 
 	"github.com/holos-run/holos-console/console/deployments"
+	"github.com/holos-run/holos-console/console/folders"
 	"github.com/holos-run/holos-console/console/oidc"
 	"github.com/holos-run/holos-console/console/organizations"
 	"github.com/holos-run/holos-console/console/projects"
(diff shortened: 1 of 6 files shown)


jeffmccune:
## Rationale

### Architecture

The FolderService follows the same pattern as OrganizationService and ProjectService: a K8s namespace backed resource with annotation-based RBAC grants. Folders are stored as namespaces with:
- `console.holos.run/resource-type=folder`
- `console.holos.run/organization={root-org}` (for org-scoped listing)
- `console.holos.run/folder={name}` (for label-based name extraction, mirrors org/project pattern)
- `console.holos.run/parent={parent-namespace}` (the hierarchy link)

### Depth enforcement

Depth is computed at create time by walking the parent chain upward and counting folder hops until an org namespace is reached. This is O(depth) = O(3) and doesn't require traversal of descendants. Depth 4 is rejected — meaning at most 3 folder levels sit between an org and a project.

### Default-share cascade

When creating a folder, `collectAncestorDefaultShares` walks from the parent namespace up to the org root and merges `default-share-users`/`default-share-roles` from every ancestor. The merge uses "highest role wins per principal" semantics. Explicit grants in the `CreateFolderRequest` are layered on top.

### Permission model

Folder-level access uses `rbac.CheckAccessGrants` (direct, non-cascading) — the same as organizations. This means grants on the folder namespace itself control access. The rbac walker from Phase 4 can be wired in a future phase if cascade from org → folder is needed.

## Alternatives considered

| Option | Verdict |
|--------|---------|
| Use hierarchy walker for auth | Rejected for this phase — the issue says to use direct grants (CheckAccessGrants), matching org behavior. Phase 4 walker can be layered later. |
| Store parent as annotation instead of label | Rejected — the existing walker.go already reads parent from `ns.Labels[v1alpha2.AnnotationParent]`, so storing in labels is consistent with the hierarchy walker design. |
| OrgNamespaceGetter interface on Handler | Removed — the handler can fetch any namespace directly via `k8s.GetNamespace`, making the extra interface unnecessary. |

PR #654, opened 2026-04-10
feat: migrate ProjectService to folder-aware parent pointer

jeffmccune:
## Summary

- Add `parentNs` parameter to `K8sClient.CreateProject` and `K8sClient.ListProjects` so projects can be nested under a folder namespace
- `Handler.resolveParentNS` converts `ParentType+ParentName` to a Kubernetes namespace name (mirrors the same helper in FolderService)
- `buildProject` now populates `ParentType` and `ParentName` on the proto using `Resolver.ResourceTypeFromNamespace`
- Migrate all `v1alpha1.*` and `resolver.*` constant references to `v1alpha2.*` in handler and k8s layers
- Update `organizations.ProjectLister` interface to include `parentNs` argument (caller passes `""`)

Closes: #629

## Test plan
- [x] `go test ./console/projects/... ./console/organizations/... ./console/folders/...` passes
- [x] `go test -race ./...` passes

### DIFF
--- a/console/organizations/handler.go
+++ b/console/organizations/handler.go
@@ -24,7 +24,7 @@ const auditResourceType = "organization"
 
 // ProjectLister checks for projects linked to an organization.
 type ProjectLister interface {
-	ListProjects(ctx context.Context, org string) ([]*corev1.Namespace, error)
+	ListProjects(ctx context.Context, org, parentNs string) ([]*corev1.Namespace, error)
 }
 
 // Handler implements the OrganizationService.
@@ -278,7 +278,7 @@ func (h *Handler) DeleteOrganization(
 	}
 
 	if h.projectLister != nil {
-		projects, err := h.projectLister.ListProjects(ctx, req.Msg.Name)
+		projects, err := h.projectLister.ListProjects(ctx, req.Msg.Name, "")
 		if err != nil {
 			return nil, connect.NewError(connect.CodeInternal, fmt.Errorf("checking for linked projects: %w", err))
 		}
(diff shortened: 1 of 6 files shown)


jeffmccune:
## Rationale

The key insight is that a project in v1alpha2 may live directly under a folder rather than directly under an org. This requires two changes:

1. **Storage**: `K8sClient.CreateProject` writes the `console.holos.run/parent` label when `parentNs` is non-empty — the same pattern used by `K8sClient.CreateFolder`. `K8sClient.ListProjects` filters by that label when `parentNs` is provided.

2. **Resolution**: `Handler.resolveParentNS` (identical to the same helper in FolderService) converts the proto `ParentType+ParentName` pair to a Kubernetes namespace name. `buildProject` reads the parent label back and uses `Resolver.ResourceTypeFromNamespace` to populate the proto's `ParentType` and `ParentName` fields — avoiding an extra K8s API call.

**Backwards compatibility**: `CreateProject` defaults `parentNs` to the org namespace when `ParentName` is empty, preserving existing callers. The `ListProjects` parentNs filter only applies when the caller explicitly sets `ParentType+ParentName`.

## Alternatives considered

| Option | Verdict |
|--------|---------|
| Add a `GetNamespace` + extra API call in `buildProject` to resolve parent info | Rejected — `ResourceTypeFromNamespace` resolves prefix → kind without a network call |
| Keep `ProjectLister` interface with old signature; add a wrapper method | Rejected — the interface is only used in organizations and adding a shim adds more complexity than just updating the signature |

PR #655, opened 2026-04-10
feat: migrate SecretsService to v1alpha2 with ancestor default-share cascade

jeffmccune:
## Summary

- Replace all `v1alpha1.*` constants with `v1alpha2.*` in `console/secrets/k8s.go` and test files
- Extend `ProjectGrantResolver.GetDefaultGrants` in `console/projects/resolver.go` to walk the full ancestor chain (project → folders → org) when a `resolver.Walker` is configured; merge default-share grants from all levels using highest-role-wins per principal
- Wire the walker in `console/console.go` via `projects.NewProjectGrantResolver(k8s).WithWalker(nsWalker)`
- Add `console/projects/resolver_test.go` with tests for: direct-under-org, under-folder, under-nested-folders, and project-overrides-org scenarios
- Enhance `checkAccess` comment to explain the principled separation: ancestry walk is ONLY for default-share cascade at create time, NOT for access checks on existing secrets (ADR 007)

Closes: #630

## Test plan
- [x] New tests in `console/projects/resolver_test.go` cover all required ancestor cascade scenarios
- [x] Existing `console/secrets/` tests pass with v1alpha2 constants
- [x] `go test -race ./...` passes

### DIFF
--- a/console/console.go
+++ b/console/console.go
@@ -282,9 +282,10 @@ func (s *Server) Serve(ctx context.Context) error {
 		projectsPath, projectsHTTPHandler := consolev1connect.NewProjectServiceHandler(projectsHandler, protectedInterceptors)
 		mux.Handle(projectsPath, projectsHTTPHandler)
 
-		// Secrets service with project grant fallback
+		// Secrets service with project grant fallback and ancestor default-share cascade.
 		secretsK8s := secrets.NewK8sClient(k8sClientset, nsResolver)
-		projectResolver := projects.NewProjectGrantResolver(projectsK8s)
+		nsWalker := &resolver.Walker{Client: k8sClientset, Resolver: nsResolver}
+		projectResolver := projects.NewProjectGrantResolver(projectsK8s).WithWalker(nsWalker)
 		secretsHandler := secrets.NewProjectScopedHandler(secretsK8s, projectResolver)
 		secretsPath, secretsHTTPHandler := consolev1connect.NewSecretsServiceHandler(secretsHandler, protectedInterceptors)
 		mux.Handle(secretsPath, secretsHTTPHandler)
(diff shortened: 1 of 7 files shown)


jeffmccune:
## Rationale

### Why the walker lives in ProjectGrantResolver, not the handler

The `DefaultShareResolver` interface used by `SecretsService.CreateSecret` takes a project name and returns merged default grants. Keeping the walk logic inside `ProjectGrantResolver.GetDefaultGrants` means:

1. The secrets handler interface is unchanged — `SecretsService` still only knows about `DefaultShareResolver`
2. The walk is opt-in via `WithWalker()`, so tests without K8s fixtures work as before (no walker = project-level only)
3. `console.go` creates one walker and shares it — consistent behavior across all usages

### Why highest-role-wins favors the child namespace

The merge order is project first, then parent folders, then org (child → parent). When the same principal appears in multiple ancestors, the child's grant wins if it has an equal or higher role. This mirrors the principle that closer-to-the-leaf policies are more specific and authoritative than broad org-level defaults.

### Why non-cascading access checks are preserved

Per ADR 007, organization-level grants don't cascade to secret data. `checkAccess` checks per-secret grants first, then project grants via `ProjectCascadeSecretPerms`. `PermissionSecretsRead` is intentionally absent from `ProjectCascadeSecretPerms` — even a project OWNER cannot read specific secrets without an explicit grant. This is correct for data (secrets contain credentials) vs policy (templates apply everywhere by design).

## Alternatives considered

| Option | Verdict |
|--------|---------|
| Walk in secrets handler directly | Rejected — handler would need to know about namespace hierarchy, violating its project-scoped design |
| Separate `AncestorDefaultShareResolver` interface | Rejected — adds interface proliferation with no benefit; `WithWalker` opt-in achieves the same goal more cleanly |
| Add walker to `K8sClient` in projects package | Rejected — the resolver is the right place since it already implements the interface contract |

PR #656, opened 2026-04-10
feat: unified TemplateService replacing DeploymentTemplateService and OrgTemplateService

jeffmccune:
## Summary
- Collapse `console/org_templates/` (`OrgTemplateService`) and the project-scoped `console/templates/` (`DeploymentTemplateService`) into a single scope-aware `TemplateService` per ADR 021 Decision 1
- Add v1alpha2 constants: `ResourceTypeTemplate`, `LabelTemplateScope`, `AnnotationLinkedTemplates`, and scope value strings
- New `TemplateScopeRef{scope, scope_name}` discriminator in all Template RPCs; `LinkedTemplateRef{scope, scope_name, name}` replacing flat linked-names string array
- `MandatoryTemplateApplier` now walks full ancestor chain (org + folder ancestors) for project creation
- New `FolderGrantResolver` in `console/folders/resolver.go` wired into template access checks
- Deleted `console/org_templates/` package (all logic merged into `console/templates/`)

Closes: #631

## Test plan
- [x] `go test ./console/...` passes
- [ ] CI checks pass


jeffmccune:
## Rationale

ADR 021 Decision 1 defines a single `TemplateService` with a `TemplateScopeRef` discriminator rather than separate `DeploymentTemplateService` and `OrgTemplateService`. This PR implements that decision by collapsing both services.

### Why a single service?

The separate services were an artifact of v1alpha1's two-level model (org + project). With v1alpha2 adding folder-level templates, a third service would be needed. Instead, a single scope-aware service reduces the number of RPCs and makes it straightforward to add new scope levels without multiplying service definitions.

### ProjectScopedResolver adapter

The `deployments` package requires a 2-arg `GetTemplate(ctx, project, name)` interface. The new unified `K8sClient.GetTemplate` takes 4 args `(ctx, scope, scopeName, name)`. Rather than changing the `deployments` package interface (which would require updating all callers), a thin `ProjectScopedResolver` adapter fixes the impedance mismatch.

### Ancestor walk order

`resolver.Walker.WalkAncestors` returns ancestors in child→parent order. `collectAncestorTemplates` reverses this to produce org→child order so that CUE unification sees more-general constraints before more-specific ones, which is the natural reading order for platform templates.

## Alternatives considered

| Option | Verdict |
|--------|---------|
| Keep separate services, add a third `FolderTemplateService` | Rejected: leads to N services for N scope levels; duplicates nearly identical CRUD logic |
| Use a single service with a scope `oneof` field per RPC | Equivalent to `TemplateScopeRef` but more verbose; `TemplateScopeRef` is cleaner |
| Change `deployments.TemplateResolver` to 4-arg | Would require updating all callers; `ProjectScopedResolver` adapter is narrower change |

PR #657, opened 2026-04-10
feat: ancestor template walk and PlatformInput.Folders in deployment renderer

### LINKED ISSUE #632: feat(deployments): render deployments with full ancestor template walk
## Parent Issue

Part of #622

## Goal

Update `console/deployments/` to render deployments using the unified TemplateService from Phase 9. The renderer walks the project's ancestor chain, collects templates from all scopes, and unifies them per ADR 016 Decision 8 (project-level source contributes only `projectResources`; folder/org contribute both). Also includes the proto field rename from `cue_user_input` to `cue_project_input` in the render preview response.

## Acceptance Criteria

- [ ] `console/deployments/render.go`:
  - `CueRenderer.RenderWithOrgTemplates` renamed to `CueRenderer.RenderWithAncestorTemplates` (or similar — clean name, no `Org` in it).
  - Takes a slice of `AncestorTemplate { Scope, ScopeName, CUE, ReadPlatformResources bool }` ordered org → folders → project.
  - `ReadPlatformResources` is `true` for org and folder scopes, `false` for project scope (enforcing ADR 016 Decision 8).
  - `evaluateStructured` uses each source's `ReadPlatformResources` flag to decide whether to read `platformResources` from that source's contribution — OR we continue to concatenate sources and read `platformResources` only from the unified tree (simpler; project-level sources define under `platformResources` but the renderer still reads the field from the unified tree because CUE unification merges them). Pick the simpler approach that correctly enforces the ADR 016 boundary and document the choice.
  - `PlatformInput` gains the `Folders []string` field from Phase 1.
- [ ] `console/deployments/handler.go`:
  - `CreateDeployment` calls `templates.ListAncestorTemplates` (Phase 9) to get the full render set.
  - `UpdateDeployment` does the same.
  - `GetDeploymentRenderPreview` returns `cue_project_input` (renamed from `cue_user_input`) and populates `PlatformInput.Folders` correctly.
- [ ] `console/deployments/apply.go` — allowed kinds list unchanged; may need updating if folders introduce new kinds (discuss in ADR 020/021).
- [ ] Tests cover:
  - Deploy a project with no ancestor templates — renders from project template only.
  - Deploy a project with an org mandatory template — template is unified and produces platform resources.
  - Deploy a project with a folder mandatory template and a project template — both unified.
  - Deploy a project linking an org template explicitly — linked template unified.
  - Deploy a project under folder-of-folder linking a grandparent template.
  - Project template defining under `platformResources` has those fields IGNORED (ADR 016 Decision 8).
  - Rollback on CUE error — partial K8s state is cleaned up (existing behavior preserved).
- [ ] `RenderDeploymentTemplate` RPC updated to accept `LinkedTemplateRef` objects from Phase 2.

### Files to read first

- `console/deployments/render.go` — current rendering pipeline (`Render`, `RenderWithOrgTemplates`, `evaluate`, `evaluateStructured`)
- `console/deployments/handler.go:250-500` — CreateDeployment / UpdateDeployment flow
- `console/deployments/apply.go` — allowed kinds + Reconcile
- ADR 016 Decision 6, 8 — renderer read boundaries
- ADR 021 (#621) — render set formula

### Enforcement of `platformResources` boundary

ADR 016 Decision 6 says the renderer reads `platformResources` only from templates at the folder level or above. Because CUE unification merges all sources into a single tree, the only way to enforce this without CUE-level tricks is:
- Evaluate org + folder sources separately and extract `platformResources` from that result.
- Evaluate all sources together and extract `projectResources` from the unified result.
- This requires two CUE compilations, which is cheap enough.

Alternatively, evaluate the full tree once and accept that a project template defining under `platformResources` is a no-op because the unified tree's `platformResources` is what gets applied — but this violates the ADR's hard boundary. Prefer the two-compile approach.

### cue_user_input → cue_project_input rename

This is a proto field rename that may be done in Phase 2 (the big proto rewrite) OR in this phase after the semantic rename lands. Prefer Phase 2 so that this phase only updates the handler implementation.

## Dependencies

- Depends on #TBD (Phases 1, 2, 3, 4, 9).
- Blocks Phase 11 (frontend deployment form needs the updated render preview).

jeffmccune:
## Summary
- Add `Folders []string` to `PlatformInput` so CUE templates can access `platform.folders` (ordered list from org → project)
- Rename `RenderWithOrgTemplates` → `RenderWithAncestorTemplates` to accurately reflect that it accepts org + folder ancestor sources
- Add `AncestorWalker` interface to deployment handler; wire `ProjectFolderResolver` (new in `console/projects/k8s.go`) to populate `Folders`
- Update `linkedTemplateNamesFromAnnotation` to support v1alpha2 `console.holos.run/linked-templates` format with legacy fallback
- Add `TestCueRenderer_FoldersPropagation` and `TestCueRenderer_AncestorTemplateWalk` test suites

Closes: #632

## Test plan
- [x] `go test ./...` passes
- [ ] CI checks pass

### DIFF
--- a/api/v1alpha1/input.go
+++ b/api/v1alpha1/input.go
@@ -39,6 +39,12 @@ type PlatformInput struct {
 	GatewayNamespace string `json:"gatewayNamespace" yaml:"gatewayNamespace" cue:"gatewayNamespace"`
 	// Organization is the parent organization name.
 	Organization string `json:"organization"     yaml:"organization"     cue:"organization"`
+	// Folders is the ordered list of folder names in the ancestor chain from
+	// the organization down to (but not including) the project. The first
+	// element is the immediate child of the organization; the last element is
+	// the immediate parent of the project. Empty when the project has no folder
+	// ancestors. Available in CUE templates as platform.folders.
+	Folders []string `json:"folders,omitempty" yaml:"folders,omitempty" cue:"folders?"`
 	// Claims carries the OIDC ID token claims of the authenticated user.
 	Claims Claims `json:"claims"           yaml:"claims"           cue:"claims"`
 }
--- a/console/console.go
+++ b/console/console.go
@@ -322,9 +322,11 @@ func (s *Server) Serve(ctx context.Context) error {
 		if dynamicClient != nil {
 			deploymentsApplier = deployments.NewApplier(dynamicClient)
 		}
+		projectFolderResolver := projects.NewProjectFolderResolver(projectsK8s, nsWalker)
 		deploymentsHandler := deployments.NewHandler(deploymentsK8s, projectResolver, settingsK8s, templates.NewProjectScopedResolver(templatesK8s), &deployments.CueRenderer{}, deploymentsApplier).
 			WithOrgProvider(projectsK8s).
-			WithOrgTemplateProvider(templatesK8s)
+			WithOrgTemplateProvider(templatesK8s).
+			WithAncestorWalker(projectFolderResolver)
 		deploymentsPath, deploymentsHTTPHandler := consolev1connect.NewDeploymentServiceHandler(deploymentsHandler, protectedInterceptors)
 		mux.Handle(deploymentsPath, deploymentsHTTPHandler)
 	} else {
--- a/console/deployments/handler.go
+++ b/console/deployments/handler.go
@@ -16,6 +16,7 @@ import (
 	"sigs.k8s.io/yaml"
 
 	v1alpha1 "github.com/holos-run/holos-console/api/v1alpha1"
+	v1alpha2 "github.com/holos-run/holos-console/api/v1alpha2"
 	"github.com/holos-run/holos-console/console/rbac"
 	"github.com/holos-run/holos-console/console/rpc"
 	consolev1 "github.com/holos-run/holos-console/gen/holos/console/v1"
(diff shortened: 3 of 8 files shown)


jeffmccune:
## Rationale

### Why rename `RenderWithOrgTemplates` → `RenderWithAncestorTemplates`?

The original name implied only org-level templates. With folder support, templates come from the full ancestor chain (org + folders). The rename makes the interface self-documenting.

### Why add `AncestorWalker` as a separate interface?

The deployment handler already has `OrgProvider` (org name resolution) and `OrgTemplateProvider` (template source fetching). `AncestorWalker` is a third concern: populating `platform.folders` in `PlatformInput`. Keeping these as separate interfaces preserves testability — tests for folder propagation need only provide a stub `AncestorWalker`, not full org resolution.

### Why `ProjectFolderResolver` in `console/projects/k8s.go`?

The projects package already owns the walker wiring (via `ProjectGrantResolver.WithWalker`). Adding `ProjectFolderResolver` there avoids introducing a new package or circular imports, and it naturally reuses the existing walker infrastructure.

### `linkedTemplateNamesFromAnnotation` legacy fallback

The v1alpha1 `linked-org-templates` annotation stored only template names. v1alpha2 `linked-templates` stores full `{scope, scope_name, name}` refs. The new function extracts only the `name` field for compatibility with the existing `OrgTemplateProvider.ListOrgTemplateSourcesForRender` interface (which takes `[]string`). The full ref is not yet used at render time; that will come in a later phase when the render set formula is scope-aware.

## Alternatives considered

| Option | Verdict |
|--------|---------|
| Keep `RenderWithOrgTemplates` name, add `RenderWithFolderTemplates` | Rejected: two nearly identical methods; cleaner to have one |
| Populate `Folders` by modifying `OrgProvider` to also return folders | Rejected: changes too many interfaces at once; `AncestorWalker` is narrower |
| Pass full `LinkedTemplateRef` slice to `ListOrgTemplateSourcesForRender` | Deferred to a later phase — current interface change is minimal |

PR #661, opened 2026-04-10
refactor(proto): rename cue_user_input to cue_project_input in deployment render preview

jeffmccune:
## Summary
- Renames Go handler variable `cueUserInput` → `cueProjectInput` in `console/deployments/handler.go` to match the proto field name
- Updates stale error message strings in `handler_test.go` from `cue_user_input` → `cue_project_input`
- Renames the frontend unit test mock key `cueUserInput` → `cueProjectInput` in the deployment detail test
- Removes historical "Replaces cue_user_input from v1alpha1" migration notes from proto comments in `deployments.proto` and `templates.proto` since the migration is complete
- Runs `make generate` to regenerate TypeScript `.d.ts` and Go `.pb.go` files from the updated proto comments

Closes: #637

## Test plan
- [x] `rg 'cue_user_input|cueUserInput'` across proto/Go/TypeScript returns zero hits
- [x] `make generate` passes
- [x] `make test` passes (42 test files, 645 UI tests, all Go packages green)

### DIFF
--- a/console/deployments/handler.go
+++ b/console/deployments/handler.go
@@ -728,7 +728,7 @@ func (h *Handler) GetDeploymentRenderPreview(
 		return nil, connect.NewError(connect.CodeInternal, fmt.Errorf("encoding project input: %w", err))
 	}
 	cuePlatformInput := fmt.Sprintf("platform: %s", string(platformJSON))
-	cueUserInput := fmt.Sprintf("input: %s", string(projectJSON))
+	cueProjectInput := fmt.Sprintf("input: %s", string(projectJSON))
 
 	// Render the template to produce YAML and JSON output, including linked
 	// platform templates (ADR 019). Uses renderResources so the same linking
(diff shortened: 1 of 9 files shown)


jeffmccune:
## Rationale

This is a pure cleanup rename. The proto field `cue_project_input` was already in place (field number 3 in `GetDeploymentRenderPreviewResponse`), but two implementation artifacts were left over from the pre-split naming convention:

1. The Go handler variable was still named `cueUserInput` — renamed to `cueProjectInput` for consistency with the proto field.
2. Test error strings still said `"expected non-empty cue_user_input"` and `"expected cue_user_input to contain ..."` — updated to use the current field name.
3. The frontend unit test mock object had key `cueUserInput` — updated to `cueProjectInput` to match the generated TypeScript type.
4. Proto comments on both `deployments.proto` and `templates.proto` contained historical migration notes ("Replaces cue_user_input from v1alpha1…") — removed since the migration is complete and the notes are no longer informative.

## Alternatives considered

| Option | Verdict |
|--------|---------|
| Keep variable name as `cueUserInput` | Rejected — creates confusion between the Go variable and the proto field name; the acceptance criteria explicitly requires zero hits for `cue_user_input\|cueUserInput` |
| Only fix the Go code, leave test strings | Rejected — stale test error messages are misleading; they should name the actual field being checked |
| Add a proto alias / reserved field for backward compatibility | Rejected — this code is not yet released (per AGENTS.md); backward compat is not required |

PR #667, opened 2026-04-10
refactor(frontend): remove deployment-templates.ts and org-templates.ts compat shims

jeffmccune:
## Summary
- Expand `templates.ts` with the full unified hook set: `useListTemplates`, `useGetTemplate`, `useCreateTemplate`, `useUpdateTemplate`, `useDeleteTemplate`, `useCloneTemplate`, `useListLinkableTemplates`, `useRenderTemplate`, plus scope helpers `makeScope`/`makeOrgScope`/`makeProjectScope`
- Update `CueTemplateEditor` to accept `scope: TemplateScopeRef` and call `useRenderTemplate` internally (previously required caller to inject a render hook function, which violated Rules of Hooks when callers tried to pass closures)
- Migrate all route components from `@/queries/deployment-templates` and `@/queries/org-templates` to `@/queries/templates`
- Remove dead linked-template UI from the project template create form (`new.tsx`) — the `linkedOrgTemplates` state was computed but never passed to `createMutation`
- Delete `deployment-templates.ts` and `org-templates.ts` (backward-compat shims, now unused)
- Update all test files to mock `@/queries/templates` instead of the removed shims

Closes: #643

## Test plan
- [x] `make test-ui` — 638 tests pass
- [x] `make generate` — TypeScript build succeeds
- [x] `make test-go` — all Go tests pass


jeffmccune:
## Rationale

### Why migrate from compat shims to a unified hook set?

`deployment-templates.ts` and `org-templates.ts` were introduced as backward-compat wrappers around the new `TemplateService` RPC (unified in ADR 021). They allowed incremental migration of callers, but perpetuated two parallel hook namespaces (`useListDeploymentTemplates` vs `useListTemplates`, etc.) with identical semantics. Keeping both created ongoing confusion about which to use for new code.

### Why change CueTemplateEditor to accept `scope` instead of `useRenderFn`?

The original design injected a `useRenderFn: (cueTemplate, cueInput, enabled, platformInput) => ...` prop so callers could supply any render hook. The new `useRenderTemplate` hook takes `scope` as its first argument. The natural migration would be to have callers pass a closure like `(t, i, e, p) => useRenderTemplate(scope, t, i, e, p)`.

The problem: that closure calls a hook (`useRenderTemplate`) inside a callback, which violates React's Rules of Hooks (hooks must be called unconditionally at the top level of a component or custom hook). Passing `scope: TemplateScopeRef` and calling `useRenderTemplate(scope, ...)` inside `CueTemplateEditor` directly is both simpler and correct.

### Why remove the linked-template checkboxes from the create form?

Inspection revealed `linkedOrgTemplates` was gathered from checkboxes but then `useCreateDeploymentTemplate` always passed `linkedTemplates: []` — the selection was never sent to the server. This was dead UI. The issue explicitly called out removing dead code, and removing it also reduces form complexity (the linking workflow belongs on the template detail page after creation).

## Alternatives considered

| Option | Verdict |
|--------|---------|
| Keep compat shims, add re-exports | Rejected — perpetuates dual namespaces, adds no value |
| Change `useRenderFn` prop to accept `scope` argument | Rejected — caller would still need to call a hook inside a callback, violating Rules of Hooks |
| Keep linked-template UI but fix the bug | Rejected — the issue explicitly asks to remove dead code; linking is better done post-creation on the detail page |

jeffmccune:
## Visual Verification

The org platform templates list page (the page most affected by this refactoring — switches from `useListOrgTemplates`/`useCreateOrgTemplate` to `useListTemplates`/`useCreateTemplate` with scope dispatch) renders correctly.

Org Platform Templates page

This is a refactoring PR with no user-visible UI changes. The screenshot confirms the page still renders and the unified hook is working correctly after the compat shims were removed.

Captured with `scripts/agent-dev | { eval "$(head -1)"; scripts/browser-login; agent-browser open ...; agent-browser screenshot ... }`.

PR #669, opened 2026-04-10
fix(ci): wait for k3s node registration before checking readiness

jeffmccune:
## Summary
- Fix E2E test failures on main caused by a race condition in k3s setup
- `kubectl wait --for=condition=Ready nodes --all` fails with "no matching resources found" when k3s hasn't registered any node resources yet
- Add a retry loop (up to 60s) that polls for node existence before running the readiness wait

## Test plan
- [ ] E2E CI job passes (the fix is self-verifying — the E2E job exercises the changed step)

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -78,7 +78,14 @@ jobs:
           curl -sfL https://get.k3s.io | sh -s - --disable=traefik
           sudo chmod 644 /etc/rancher/k3s/k3s.yaml
           echo "KUBECONFIG=/etc/rancher/k3s/k3s.yaml" >> $GITHUB_ENV
-          KUBECONFIG=/etc/rancher/k3s/k3s.yaml kubectl wait --for=condition=Ready nodes --all --timeout=120s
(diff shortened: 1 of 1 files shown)


jeffmccune:
## Rationale

The E2E CI job on main (commit e4f09e2) failed because k3s hadn't registered any node resources by the time `kubectl wait --for=condition=Ready nodes --all` ran. The `--timeout=120s` flag only applies to *existing* resources that haven't met the condition yet — if no nodes exist at all, `kubectl wait` exits immediately with "no matching resources found".

The fix adds a polling loop that waits for at least one node resource to appear (up to 60s) before invoking `kubectl wait`. This handles the startup race without increasing the overall timeout.

## Alternatives considered

| Option | Verdict |
|--------|---------|
| `sleep 10` before `kubectl wait` | Rejected — fragile, doesn't adapt to variable startup times |
| `--timeout=300s` on kubectl wait | Doesn't help — the error is "no matching resources found", not a timeout. `kubectl wait` exits immediately when no resources match the selector |
| Retry the entire `kubectl wait` command | Would work but is noisier; splitting the concern (wait for existence, then wait for readiness) is clearer |

PR #687, opened 2026-04-11
fix(skills): replace hardcoded codex path with portable lookup script

jeffmccune:
## Summary
- Remove hardcoded `/Users/u6136576/.nvm/versions/node/v24.12.0/bin/codex` path from `review-pr` skill
- Add `scripts/check-codex` helper that resolves the codex binary via PATH
- Update the skill to use the helper script for portable codex discovery

## Test plan
- [ ] `scripts/check-codex` exits 0 and prints `CODEX="..."` when codex is installed
- [ ] `scripts/check-codex` exits non-zero with a diagnostic when codex is not on PATH
- [ ] No hardcoded user paths remain in `.claude/skills/`

Generated with [Claude Code](https://claude.com/claude-code) · ${SLOT}

### DIFF
--- a/.claude/skills/review-pr/SKILL.md
+++ b/.claude/skills/review-pr/SKILL.md
@@ -21,16 +21,13 @@ Currently uses the OpenAI Codex CLI as the review backend. The backend is swappa
 
 ### 1. Preflight
 
-Verify codex is available:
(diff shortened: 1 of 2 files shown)


jeffmccune (commented):

PR #816, opened 2026-04-12
fix: use scope-qualified keys for linked template selection

### LINKED ISSUE #804: fix: address review findings from PR #801
## Context

PR #801 was merged with a non-critical review finding. This should be addressed in a follow-up.

### [IMPORTANT] Name-only keys for linked template selection are ambiguous
If org and folder templates share the same name, the checkbox selection uses name-only keys which could collide. Both the create page (new in PR #801) and the existing detail page share this pattern. Fix: use scope-qualified keys (e.g., `org:template-name`, `folder:template-name`) for linked template selection state on both pages.

## Source

Review output: `tmp/review-pr/pr-801/round-1.md`

Parent plan: #765

jeffmccune:
## Summary
- Extracted `linkableKey()` and `parseLinkableKey()` from the template detail page into the shared `queries/templates.ts` module so both pages use the same scope-qualified key format
- Updated the create template page to use scope-qualified composite keys (`scope/scopeName/name`) instead of name-only keys for linked template checkbox selection state
- This prevents collisions when org and folder templates share the same name -- previously checking one would affect both, and the create mutation would resolve to the wrong scope

Closes #804

## Test plan
- [x] New test: "disambiguates same-name templates across org and folder scopes" -- sets up two templates with the same name in different scopes and verifies only the selected scope is sent in the mutation
- [x] All 31 create page tests pass
- [x] All 64 detail page tests pass
- [x] All 715 UI tests pass (47 test files)
- [x] All Go tests pass
- [x] TypeScript compiles cleanly (no type errors)
- [ ] Local E2E was not run (no k3d cluster available). Relying on CI E2E check.

### DIFF
--- a/frontend/src/queries/templates.ts
+++ b/frontend/src/queries/templates.ts
@@ -15,6 +15,17 @@ import { useAuth } from '@/lib/auth'
 export type { TemplateScopeRef, LinkableTemplate, LinkedTemplateRef }
 export { TemplateScope }
 
+/** Build a composite key that uniquely identifies a linkable template across scopes. */
+export function linkableKey(scope: number | undefined, scopeName: string | undefined, name: string): string {
+  return `${scope ?? 0}/${scopeName ?? ''}/${name}`
+}
+
+/** Parse a composite key back into its constituent parts. */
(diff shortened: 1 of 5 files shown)


jeffmccune (commented):

PR #830, opened 2026-04-12
fix: use isPending instead of isSuccess for linkable templates loading state

### LINKED ISSUE #825: fix: use isPending instead of isSuccess for linkable templates loading state
## Context

PR #824 was merged with non-critical review findings that remain after round 2 fixes. These should be addressed in a follow-up.

## Findings

Both findings are the same pattern — using TanStack Query's `isSuccess` (aliased as `linkableReady`) instead of `isPending` to gate the loading state. If the query errors (including background refetch failures), users see perpetual "Loading..." instead of the empty state or cached data.

1. **`frontend/src/routes/_authenticated/projects/$projectName/templates/new.tsx:241`** — Change `isSuccess` to `!isPending` for the loading guard so that error states fall through to the empty state rather than stuck on "Loading..."

2. **`frontend/src/routes/_authenticated/projects/$projectName/templates/$templateName.tsx:55`** — Same fix: use `!isPending` instead of `isSuccess`

## Source

Review output: `tmp/review-pr/pr-824/round-2.md`

Parent plan: #819

jeffmccune:
## Summary
- Changed the loading guard for linkable templates from `isSuccess` to `isPending` in both the Create Template page (`new.tsx`) and the Template Detail page (`$templateName.tsx`)
- When the query errors (including background refetch failures), the UI now falls through to the empty state instead of showing "Loading..." indefinitely
- Added regression tests for both components covering the pending and error states
- Updated all test mocks across 3 test files to use `isPending: false` instead of `isSuccess: true`

Closes #825

## Test plan
- [x] All 818 UI tests pass (`make test-ui`)
- [x] New tests verify pending state shows loading indicator
- [x] New tests verify error state does NOT show loading indicator (falls through to empty state)
- [ ] Local E2E was not run (no k3d cluster available). Relying on CI E2E check.

### DIFF
--- a/frontend/src/routes/_authenticated/projects/$projectName/templates/$templateName.tsx
+++ b/frontend/src/routes/_authenticated/projects/$projectName/templates/$templateName.tsx
@@ -52,7 +52,7 @@ export function DeploymentTemplateDetailPage({ projectName: propProjectName, tem
   const scope = makeProjectScope(projectName)
   const { data: template, isPending, error } = useGetTemplate(scope, templateName)
   const { data: project } = useGetProject(projectName)
-  const { data: linkableTemplates = [], isSuccess: linkableReady } = useListLinkableTemplates(scope)
+  const { data: linkableTemplates = [], isPending: linkablePending } = useListLinkableTemplates(scope)
   const updateMutation = useUpdateTemplate(scope, templateName)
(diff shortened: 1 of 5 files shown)


jeffmccune (commented):

Issue #925, opened 2026-04-13
feat(templates): reliable form pre-fill via defaults RPC on template select


jeffmccune:
## Problem

When a user visits `/projects/:project/deployments/new` and selects a project template (e.g. `httpbin-v1`), the Create Deployment form fields **do not auto-populate**. The browser Network tab shows no RPC call on selection. Defaults currently ride along in `ListTemplates` as `Template.defaults`, populated from `ExtractDefaults()` which only reads a top-level `defaults:` CUE block (ADR 018/025). Because our shipped example templates (e.g. `console/templates/example_httpbin.cue`) express defaults only via CUE inline default syntax on `input` (`image: string | *"ghcr.io/…"`, `tag: string | *"2.21.0"`, `port: >0 & <=65535 | *8080`), they yield no extractable defaults block, so the UI receives an empty `TemplateDefaults` and pre-fill silently no-ops.

This behavior has been planned and re-implemented multiple times in recent days without sticking. This plan establishes a single authoritative ADR, a dedicated RPC that the UI calls on every template selection, deterministic pristine/load-defaults behavior in the form, and fully defaulted example templates — so the regression cannot recur.

## Acceptance Criteria

- [ ] New ADR (027) defines the authoritative behavior for template defaults pre-fill. ADR 018 and ADR 025 are marked **Superseded by ADR 027**.
- [ ] A dedicated RPC, `TemplateService.GetTemplateDefaults`, returns a `TemplateDefaults` for `(scope, template_name)`. Backwards compatible.
- [ ] Frontend calls `GetTemplateDefaults` every time the template selection changes AND every time the user clicks a new **Load defaults** button.
- [ ] Pristine form → selecting any template replaces all defaultable fields with the response.
- [ ] Dirty form → selecting a different template does NOT overwrite fields; only **Load defaults** force-replaces and resets pristine.
- [ ] Every shipped example template includes a top-level `defaults: #ProjectInput & { … }` block with concrete values.
- [ ] `docs/agents/guardrail-template-defaults.md` rewritten to match ADR 027.
- [ ] Backend + frontend unit tests cover every rule; E2E reserved for final cleanup sweep.

## Implementation Plan

This issue tracks the full implementation. Sub-issues cover each phase:

- [ ] #926 — docs(adr): ADR 027 authoritative template defaults pre-fill behavior
- [ ] #927 — feat(proto): add GetTemplateDefaults RPC to TemplateService
[...]

PR #943, opened 2026-04-13
feat(ui): call GetTemplateDefaults on template change and add Load defaults button

### LINKED ISSUE #929: feat(ui): call GetTemplateDefaults on template change and add Load defaults button
## Parent Issue

Part of #925

## Goal

Rewrite the Create Deployment form's template-change behavior so selecting a template issues an explicit \`GetTemplateDefaults\` RPC and pre-fills the form under ADR 027's pristine rules. Add a **Load defaults** button that force-overwrites, and cover every rule with Vitest unit tests against mocked query hooks.

## Acceptance Criteria

- [ ] \`frontend/src/queries/templates.ts\` exports a \`useGetTemplateDefaults({ scope, name }, { enabled })\` hook wrapping the generated ConnectRPC Query client for \`TemplateService.GetTemplateDefaults\`. The hook is not eagerly called when \`name\` is empty.
- [ ] \`frontend/src/routes/_authenticated/projects/$projectName/deployments/new.tsx\` is reworked to:
  - Track \`isPristine: boolean\` state, initialized \`true\`. Every defaultable field's change handler sets it to \`false\`.
  - On template change: set the \`template\` state, trigger \`GetTemplateDefaults\` via the query hook (refetch if already cached for that key), and in the success handler, if \`isPristine === true\`, overwrite every defaultable field from the response and leave \`isPristine\` as \`true\`; otherwise leave fields untouched.
  - Add a **Load defaults** button adjacent to the Template Combobox (disabled when no template is selected or the RPC is in flight). Clicking it refetches \`GetTemplateDefaults\` and unconditionally overwrites all defaultable fields from the response, then sets \`isPristine\` to \`true\`.
  - Remove the current behavior that reads \`templates.find(...).defaults\` from \`useListTemplates\` for form pre-fill. (The list hook still renders the combobox options.)
  - Defaultable fields: \`displayName\`, \`name\` (slug — derived from \`displayName\` when \`displayName\` is in defaults; otherwise set directly from \`defaults.name\`), \`description\`, \`image\`, \`tag\`, \`port\`, \`command\`, \`args\`, \`env\`.
- [ ] The \`port\` field only pre-fills when the response carries a nonzero value; when zero, it falls back to the current form value (avoids clobbering a user-friendly default of 8080 with 0).
- [ ] Visual: the Load defaults button uses the existing \`Button\` component with \`variant=\"outline\"\`, placed to the right of the Combobox. Label is exactly \"Load defaults\".
- [ ] Vitest + React Testing Library tests in \`frontend/src/routes/_authenticated/projects/$projectName/deployments/-new.test.tsx\` cover:
  - Pristine form → selecting httpbin-v1 calls \`GetTemplateDefaults\` once and fills every defaultable field from the mocked response.
  - Pristine form → switching from templateA to templateB calls the RPC again and replaces fields with templateB's defaults.
  - User edits \`image\` (form now dirty) → selecting templateB calls the RPC but fields are NOT overwritten.
  - Dirty form → clicking **Load defaults** calls the RPC and overwrites every defaultable field; subsequently editing a field and switching templates again no longer overwrites (pristine was reset and then re-dirtied).
  - No template selected → **Load defaults** is disabled.
  - Response omits \`port\` (0) → the form keeps its prior \`port\` value.
- [ ] Tests pass: \`make test\` (frontend + backend).

## Implementation Notes

- Do not read \`Template.defaults\` from the list response for form pre-fill. ADR 027 names the explicit RPC as the sole source so the behavior is mockable in one place.
- Use the generated ConnectRPC Query helpers; mock via \`vi.mock('@/queries/templates')\` and assert the spy count to prove the RPC fired on every template change.
- \`displayName\` pre-fill policy: if \`defaults.name\` is set, pre-fill \`displayName\` with it and derive \`name\` (slug) from it via the existing \`slugify\` helper — preserves today's UX.
- When applying defaults, do it inside a single state update batch (e.g. React 19 automatic batching is fine). After applying, explicitly set \`isPristine\` to \`true\` so the act of programmatic pre-fill does not mark the form dirty.
- The pristine flag is scoped to the form instance — navigating away and back resets it.
- Keep the existing form validation untouched.

### Files to modify

- \`frontend/src/routes/_authenticated/projects/$projectName/deployments/new.tsx\`
- \`frontend/src/queries/templates.ts\` (add \`useGetTemplateDefaults\` hook)
- \`frontend/src/routes/_authenticated/projects/$projectName/deployments/-new.test.tsx\` (expand cases)

### Testing approach

Vitest + RTL with \`vi.mock\` on \`@/queries/templates\` returning deterministic \`TemplateDefaults\` payloads. No E2E — the RPC contract is covered by backend tests (phase 3) and the UI behavior is covered here.

### Dependencies

Depends on #928 (handler and generated types available).

jeffmccune:
## Summary

Implements ADR 027 pre-fill behavior in the Create Deployment form (plan #925 phase 4).

- Adds `useGetTemplateDefaults({ scope, name }, { enabled })` hook in `frontend/src/queries/templates.ts` wrapping the generated ConnectRPC client for `TemplateService.GetTemplateDefaults`. The hook is gated on non-empty `name` so it never fires eagerly.
- Rewrites `frontend/src/routes/_authenticated/projects/$projectName/deployments/new.tsx` to:
  - Track `isPristine` state (starts `true`, flipped to `false` by every defaultable-field edit).
  - Fire `GetTemplateDefaults` on every template selection via the query hook (key includes template name).
  - Apply the response when `isPristine` is `true`; otherwise preserve in-progress edits.
  - Add a **Load defaults** button (shadcn `Button`, `variant="outline"`) adjacent to the Template Combobox; disabled when no template is selected or RPC is in flight; click refetches and unconditionally overwrites every defaultable field, then resets `isPristine`.
  - Preserve the form's `port` value when the response carries `port == 0` (avoids clobbering the user-friendly 8080 default).
  - Stop reading `Template.defaults` from `useListTemplates` for pre-fill — the list hook renders only Combobox options.
- Exports a `DEFAULTABLE_FIELDS` constant (the ADR 027 §6 enumeration) per the guardrail's "one code-level copy" rule.
- Updates `-new.test.tsx` to mock `useGetTemplateDefaults` with a per-template defaults map and cover all ADR 027 cases: pristine fill, pristine template-switch replacement, dirty-form non-overwrite, Load defaults force-overwrite + pristine reset + re-dirty semantics, disabled state with no template, and `port==0` preservation.

Closes #929

## Test plan

- [x] `make test` — 912/912 passed (Go + UI).
- [x] New Vitest file passes in isolation — 40/40.
- [x] `make generate` produces no diff.
- [ ] CI: Unit Tests, Lint, E2E.

Agent: agent-3 (`/home/jeff/workspace/holos-run/holos-console-agent-3`)


jeffmccune (commented):
Codex review round 1: 2 findings (0 critical, 2 important, 0 style). Verdict: REQUEST_CHANGES (posted as COMMENT because GitHub does not allow requesting changes on one's own PR).

Note: Codex flagged a P1 "generated code" finding against `frontend/src/gen/holos/console/v1/templates_pb.js`, but that file is NOT in this PR's diff (only 3 files modified, all hand-written). Treating as a false positive.

Legitimate findings (both IMPORTANT, error-handling category):

### [IMPORTANT] Pristine auto-fill treats GetTemplateDefaults failure as empty defaults
- **File**: frontend/src/routes/_authenticated/projects/$projectName/deployments/new.tsx:146-162
- **Category**: error-handling / reliability
- **Issue**: If `GetTemplateDefaults` fails for a newly selected template, React Query leaves `fetchedDefaults` undefined and `defaultsFetching` becomes false, so the effect falls through to `applyDefaults(fetchedDefaults)` and treats the error as "template has no defaults". On a pristine form this clears defaultable fields and marks the template as already applied, with no signal of the failure.
- **Fix**: Gate the apply path on query `isSuccess` (or non-error state) rather than just `!isFetching`; surface the error via a toast/inline message and leave the form untouched.

### [IMPORTANT] Load defaults click treats refetch failure as success
- **File**: frontend/src/routes/_authenticated/projects/$projectName/deployments/new.tsx:175-183
- **Category**: error-handling / reliability
- **Issue**: `refetchDefaults()` resolves to a query result even when the RPC errors. The handler applies `result.data` and resets `isPristine` unconditionally, which can overwrite the form with `undefined`/stale defaults and silently mark the form pristine when the reload actually failed.
- **Fix**: Check `result.isError` / `result.status === 'error'` before applying; surface the error and leave the user's edits intact.

jeffmccune (commented):
Codex review round 2: 1 finding (1 important reliability)

### [IMPORTANT] Stale Load defaults response can apply to wrong template
- **File**: frontend/src/routes/_authenticated/projects/$projectName/deployments/new.tsx:185-204
- **Category**: reliability
- **Issue**: If the user clicks **Load defaults** and then switches the template before that refetch resolves, the handler still applies the old response unconditionally because it closes over the previous `template` value. The form can be overwritten with defaults for template A while the UI now shows template B selected.
- **Fix**: Gate the result on the current selection (compare requested template name against current `template` value before applying), or cancel/ignore the in-flight request when `handleTemplateChange` runs.

jeffmccune:

After 2 review rounds, 1 important reliability finding remains unresolved:

- **`frontend/src/routes/_authenticated/projects/$projectName/deployments/new.tsx:185-204`** — Stale-response race in `handleLoadDefaults`: clicking Load defaults, then switching templates before the refetch resolves, applies defaults for the prior template to the new selection. Fix: gate the apply on current selection or cancel/ignore in-flight requests when `handleTemplateChange` runs.

See `tmp/review-pr/pr-943/round-2.md` for details.

jeffmccune (commented):
Codex review round 3: 1 finding (1 critical).

### [CRITICAL] Re-run pristine prefill after dropping a stale Load defaults response

- **File**: frontend/src/routes/_authenticated/projects/$projectName/deployments/new.tsx:164-166
- **Category**: reliability (race condition)
- **Issue**: If the form is still pristine and the user switches templates while a previous Load defaults refetch is in flight, the new template's query can resolve while `loadDefaultsPendingRef` is still `true`. The prefill effect returns early in that case, and because `loadDefaultsPendingRef` is only a ref, clearing it in the `finally` block does not trigger the effect to re-run. Result: template B is selected but template A's defaults remain (or the form stays empty), and the wrong values can be submitted.
- **Fix**: After clearing `loadDefaultsPendingRef` in the `finally` block when the stale-response branch fires, force the pristine-prefill path to re-evaluate for the now-current template. Options: (a) store the pending flag in React state instead of a ref so the effect depends on it, (b) after the stale drop, reset `lastAppliedTemplateRef.current = ''` and bump a trigger state value so the effect re-runs with the still-cached fetchedDefaults, or (c) call `refetchDefaults()` explicitly for the current `templateRef.current` inside the stale branch.

(Codex also noted a P2 in `console/deployments/statuscache/cache.go`, but that file is not in this PR's diff and is out of scope.)

jeffmccune (commented):
Codex review round 4: 1 valid style-ish finding, 1 false positive.

**Findings:**

**[P2] Respect `displayName` defaults separately from slug defaults** — false positive. The `TemplateDefaults` proto message (proto/holos/console/v1/templates.proto:128) has no separate `displayName` field — only `name` and `description`. Issue #929's implementation notes explicitly state: "if defaults.name is set, pre-fill displayName with it and derive name (slug) from it via the existing slugify helper — preserves today's UX." The code at new.tsx:115-121 implements exactly that contract. No fix needed unless the proto schema changes.

**[P3/STYLE] Clear stale Load-defaults error banner on successful retry** — new.tsx:222-227. When `Load defaults` fails once and the user retries successfully, the manual `error` string set on the error path is never cleared on success. The alert can persist after defaults are re-applied.
Fix: call `setError(null)` at the top of `handleLoadDefaults` (or on the success branch before `applyDefaults`) so stale failure banners clear on retry.

No CRITICAL or IMPORTANT findings. The secondary race fix (lastAppliedTemplateRef guard + loadDefaultsPending as state, templateRef stale-response guard) and CI workflow changes look correct.

Verdict: COMMENT (not blocking).

PR #945, opened 2026-04-13
fix(deployments): honor rollout convergence and prefer live replica counts

### LINKED ISSUE #941: fix(deployments): address codex review findings from PR #937
## Context

Codex review of PR #937 surfaced two backend findings that pre-date this PR's changes (they were introduced by merged sibling PRs #924/#932 in parent plan #912). Capturing them as a follow-up so the frontend PR could merge cleanly.

### [IMPORTANT] Keep rollout-in-progress deployments out of RUNNING phase

`console/deployments/statuscache/cache.go:181-185`

Phase derivation marks a deployment `RUNNING` whenever `Available=True` and `ready == desired`, but Kubernetes can satisfy both while a rollout is still in progress (e.g., `updated_replicas < desired` or `observed_generation < metadata.generation`). In that state the listing shows `Running N/N` for rows that are still reconciling, defeating the `status_summary` accuracy goal for active rollouts.

Fix: treat `observed_generation < metadata.generation` or `updated_replicas < desired` as still-reconciling (PROGRESSING or UNSPECIFIED).

### [IMPORTANT] Prefer live replica counts over cached summary in GetDeploymentStatus

`console/deployments/status.go:205-208`

`GetDeploymentStatus` already fetches the live `apps/v1.Deployment`, but these lines overwrite its replica counts with the informer-cache snapshot whenever any summary entry exists. The informer is eventually consistent, so the detail page can show stale `ready/desired/available` during a rollout or immediately after an update even though this RPC has fresher data in `dep.Status`.

Fix: prefer `dep.Status.*` over cache for the replica counts on the detail path; keep cache for derived phase/message only.

## Source

Review: \`tmp/review-pr/pr-937/round-2.md\`
Parent plan: #912

jeffmccune:
## Summary
- Phase derivation in `statuscache` now requires rollout convergence (`observedGeneration >= metadata.generation` AND `updatedReplicas >= desired`) before reporting `RUNNING`; otherwise the phase stays `PENDING` to reflect in-progress reconciliation.
- `GetDeploymentStatus` now sources `ready/desired/available` replica counts exclusively from the live `apps/v1.Deployment.Status`, avoiding stale overrides from the eventually-consistent informer cache. The cached summary is still surfaced for derived phase/message display.
- Adds regression tests for both findings.

Closes #941

## Test plan
- [x] `go test ./console/deployments/statuscache/ ./console/deployments/` passes
- [x] `make test` passes
- [x] `make generate` produces no diff
- [x] `make lint` — no new findings in touched files (27 pre-existing issues on main unchanged)

### DIFF
--- a/console/console.go
+++ b/console/console.go
@@ -350,18 +350,14 @@ func (s *Server) Serve(ctx context.Context) error {
 		// console-managed apps/v1.Deployment resources via the managed-by
 		// label. The informer lifecycle is tied to the server context so it
 		// stops cleanly on shutdown. Cache misses are treated as "no data
-		// yet" by the handler, so a failure to start the cache (transient
-		// API latency, missing watch RBAC, etc.) must not block startup:
-		// log and fall through to a no-op cache so the console continues to
-		// serve, just without status summaries until the operator fixes the
-		// underlying problem and restarts.
-		deploymentStatusCache, err := statuscache.New(ctx, k8sClientset)
-		if err != nil {
-			slog.WarnContext(ctx, "deployment status cache unavailable, deployments will report UNSPECIFIED status",
-				slog.Any("error", err),
-			)
-			deploymentStatusCache = statuscache.NewNop()
-		}
+		// yet" by the handler. New is non-blocking: it starts the informer
+		// in the background and reports misses until the initial sync
+		// completes, so a transiently unreachable API server or a missing
+		// watch RBAC rule never delays startup. If the initial sync does
+		// not complete within statuscache's internal timeout, the package
+		// itself cancels the reflector to avoid leaking LIST/WATCH retry
+		// goroutines and logs a warning.
+		deploymentStatusCache := statuscache.New(ctx, k8sClientset)
 		deploymentsHandler := deployments.NewHandler(deploymentsK8s, projectResolver, settingsK8s, templates.NewProjectScopedResolver(templatesK8s), &deployments.CueRenderer{}, deploymentsApplier).
 			WithAncestorWalker(projectFolderResolver).
 			WithAncestorTemplateProvider(ancestorTemplateResolver).
--- a/console/deployments/status.go
+++ b/console/deployments/status.go
@@ -188,28 +188,22 @@ func (h *Handler) GetDeploymentStatus(
 		slog.String("sub", claims.Sub),
 	)
 
-	// Source phase-summary from the informer cache so all status RPCs share
-	// one derivation path. On a cache miss, summary is nil but the scalar
-	// replica fields still fall back to the live apps/v1.Deployment.Status
-	// values we already fetched above — otherwise a cold informer (right
-	// after startup, after the initial sync timeout, or without watch RBAC)
-	// would render healthy deployments as 0/0 ready even though the detail
-	// page has real pod data to show alongside.
+	// Replica scalars always come from the live apps/v1.Deployment.Status we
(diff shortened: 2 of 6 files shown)


jeffmccune (commented):
Codex review round 1: 2 findings (0 CRITICAL, 2 IMPORTANT, 0 STYLE)

The rollout-status logic itself looks covered, but the new shared status cache has failure-path bugs during initialization: it can delay startup and leak informer activity after falling back to the no-op cache. Those issues make the patch unsafe in the misconfigured or degraded environments this fallback is intended to support.

### [IMPORTANT] Avoid blocking startup on optional status cache sync
- **File**: console/deployments/statuscache/cache.go:103-110
- **Issue**: When the Deployment watch cannot establish (e.g., missing `list/watch deployments` RBAC or a temporarily unavailable API server), this path still waits up to `initialSyncTimeout` before returning the error that `console.go` degrades to `NewNop()`. That turns an explicitly optional feature into a guaranteed ~10s startup/readiness delay on every restart in the failure modes the fallback is meant to tolerate.
- **Fix**: Detect non-recoverable watch errors early (via the informer's WatchErrorHandler or a short pre-flight LIST) and bail out without waiting for the full sync timeout.

### [IMPORTANT] Stop the informer when initial cache sync fails
- **File**: console/deployments/statuscache/cache.go:103
- **Issue**: If `WaitForCacheSync` times out or returns false, `factory.Start(ctx.Done())` has already launched reflector goroutines on the long-lived server context. `New` then returns an error and the caller switches to `NewNop()`, but the abandoned informer keeps retrying LIST/WATCH requests for the lifetime of the process, creating background load in exactly the fallback case this code is trying to isolate.
- **Fix**: Use a derived cancellable context for `factory.Start` and cancel it on the error path before returning, so reflector goroutines exit cleanly.

Output: tmp/review-pr/pr-945/round-1.md

jeffmccune (commented):
Codex review round 2: 1 finding (1 important)

### [IMPORTANT] Derive desired replicas from spec, not status

- **File**: console/deployments/statuscache/cache.go:214
- **Category**: acceptance-criteria / reliability
- **Issue**: On a scale-up or template rollout before any new pods exist, `status.replicas` still reflects the old/current pod count, not the target count. In that state `updatedReplicas >= desired` and `ready == desired` can both be true even though the new rollout has not converged yet, so the cache will still emit `RUNNING`/`N/N` for an in-progress deployment — the acceptance criterion this PR is supposed to fix.
- **Fix**: Base `desired` on `spec.replicas` (with Kubernetes' default of 1 when nil) instead of `status.replicas`. Add a regression test for a rollout where `observedGeneration` has advanced but no new replicas have been created yet.

jeffmccune:

After 2 review rounds, the following important finding remains unresolved:

### `console/deployments/statuscache/cache.go:214` — desired sourced from status.replicas

`desired` is sourced from `status.replicas`, which during an in-progress scale-up/rollout still reflects the old pod count. Result: `RUNNING N/N` may still be reported for unconverged rollouts, partially defeating the #941 acceptance criterion for rollout-in-progress handling.

Recommended fix: derive `desired` from `spec.replicas` (default 1 when nil) and add a regression test for rollouts where `observedGeneration` advanced but new replicas have not yet been created.

Review output: `tmp/review-pr/pr-945/round-2.md`

jeffmccune (commented):
Codex review round 3: no findings. Acceptance criteria for #941 satisfied; rollout-phase logic and live replica-count handling match spec with test coverage. LGTM.

PR #960, opened 2026-04-17
feat(console): extract output.url in render pipeline and surface via GetDeploymentRenderPreview

jeffmccune:
## Summary

Phase 2 of the HOL-543 plan (Phase 1 HOL-544 already merged).

- Add `OutputJSON *string` to `GroupedResources` and extend `populateStructuredJSON` so the render pipeline extracts the CUE `output` section through the existing `extractCuePathJSON` helper.
- In `GetDeploymentRenderPreview`, unmarshal `grouped.OutputJSON` into `*consolev1.DeploymentOutput` via a local anonymous struct and set it on the response. Malformed JSON logs a warning and leaves the field unset rather than erroring the RPC.
- Preserve the present-but-empty case: `output: {}` produces a non-nil `DeploymentOutput{Url: ""}` so the frontend decides whether to render.
- Extend `stubRenderer`/`trackingDeploymentRenderer` with an `outputJSON` field so handler tests can simulate templates with/without an `output` block.
- Table-driven tests in `render_test.go` cover absent, populated, and empty-but-present OutputJSON. Table-driven tests in `handler_test.go` cover the four Output states on the response (nil / populated url / empty url / malformed JSON).

Fixes HOL-545

## Test plan

- [x] `go test ./console/deployments/ -run TestStructuredJSONExtraction -count=1 -v` (new OutputJSON subtests pass; existing five sections pass)
- [x] `go test ./console/deployments/ -run TestHandler_GetDeploymentRenderPreview -count=1 -v` (four new Output subtests pass)
- [x] `go test ./... -count=1` (all Go packages pass)
- [x] `make test` (Go + frontend: 912 frontend tests pass; all Go packages pass)
- [x] `make vet` (clean)
- [ ] Local E2E not run; no k3d cluster in this environment. Relying on CI E2E check.

### DIFF
--- a/console/deployments/handler.go
+++ b/console/deployments/handler.go
@@ -870,12 +870,14 @@ func (h *Handler) GetDeploymentRenderPreview(
 	// Extract structured JSON fields from the grouped result if render succeeded.
 	var defaultsJSON, platformInputJSON, projectInputJSON *string
 	var platformResourcesStructJSON, projectResourcesStructJSON *string
+	var output *consolev1.DeploymentOutput
 	if grouped != nil {
 		defaultsJSON = grouped.DefaultsJSON
 		platformInputJSON = grouped.PlatformInputJSON
 		projectInputJSON = grouped.ProjectInputJSON
 		platformResourcesStructJSON = grouped.PlatformResourcesStructJSON
 		projectResourcesStructJSON = grouped.ProjectResourcesStructJSON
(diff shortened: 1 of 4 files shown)


jeffmccune (commented):
Codex review round 1: no findings. LGTM. (Posted as COMMENT because GitHub prohibits approving one's own PR.)

PR #961, opened 2026-04-17
feat(frontend): render deployment output URL on Status tab

jeffmccune:
## Summary

- Read `preview.output?.url` from the existing `useGetDeploymentRenderPreview` query on the deployment detail page.
- Render an "App URL" row at the top of the Status tab (above Replicas) when the URL is a non-empty string; anchor opens in a new tab with `rel="noopener noreferrer"`.
- Render nothing for the URL row when `output` is undefined, `output.url` is an empty string, or the preview query is still pending — keeps the Status tab stable on first load.
- Extend Vitest coverage: new describe block with four cases (URL present, URL absent, URL empty, preview pending); existing `useGetDeploymentRenderPreview` mock shape widened with optional `output` field.
- Phase 3 of HOL-543. Depends on the backend work that merged in HOL-544 (proto / schema) and HOL-545 (pipeline wiring).

Fixes HOL-546

## Test plan

- [x] `cd frontend && npx vitest run '-$deploymentName.test.tsx'` — all 71 tests pass (4 new)
- [x] `cd frontend && npm test -- --run` — 916 tests pass across 56 files
- [x] `cd frontend && npx tsc --noEmit` — clean
- [x] `cd frontend && npx eslint` on modified files — no new errors (one pre-existing warning on an unrelated useEffect)
- [ ] Manually confirm in a running console that a template publishing a URL shows the "App URL" row on the Status tab, that the link opens in a new tab, and that deployments without a URL show no row

> Local E2E was not run (no k3d cluster available). Relying on CI E2E check.

### DIFF
--- a/frontend/src/routes/_authenticated/projects/$projectName/deployments/$deploymentName.tsx
+++ b/frontend/src/routes/_authenticated/projects/$projectName/deployments/$deploymentName.tsx
@@ -36,7 +36,7 @@ import {
   TableHeader,
   TableRow,
 } from '@/components/ui/table'
-import { ArrowLeft, CheckCircle2, Copy, Info, TriangleAlert, XCircle } from 'lucide-react'
+import { ArrowLeft, CheckCircle2, Copy, ExternalLink, Info, TriangleAlert, XCircle } from 'lucide-react'
 import { Role } from '@/gen/holos/console/v1/rbac_pb'
 import type { EnvVar, Event, ContainerStatus } from '@/gen/holos/console/v1/deployments_pb'
 import { useGetDeployment, useGetDeploymentStatus, useGetDeploymentLogs, useGetDeploymentRenderPreview, useUpdateDeployment, useDeleteDeployment } from '@/queries/deployments'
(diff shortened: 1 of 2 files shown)


jeffmccune (commented):
Codex review round 1: 1 finding (1 critical, 0 important, 0 style). Posting as COMMENT because GitHub does not allow the PR author to REQUEST_CHANGES on their own PR — but this should be treated as merge-blocking.

### [CRITICAL] Validate output.url scheme before rendering as anchor href

- **File**: `frontend/src/routes/_authenticated/projects/$projectName/deployments/$deploymentName.tsx:319-321`
- **Category**: security
- **Issue**: If a template sets `output.url` to a non-HTTP(S) scheme such as `javascript:...`, this code renders it directly into an `<a href>` and clicking the new App URL row can execute script in the console UI context (XSS via unsafe URL scheme). Because `output.url` now comes from template data rather than a hardcoded constant, it must be constrained to safe URL schemes before rendering the link.
- **Fix**: Parse the URL and only render the anchor when the scheme is `http:` or `https:` (otherwise omit the row). Guard with a helper like `isSafeHttpUrl(value)` that returns true only for `http:`/`https:` protocols, and add a Vitest case that asserts a `javascript:alert(1)` URL does NOT render the App URL link.

Full review: `tmp/review-pr/pr-961/round-1.md`

PR #962, opened 2026-04-17
feat(console): cache output.url and surface Open link on deployments list

jeffmccune:
## Summary

Phase 4 of HOL-543 — completes the output.url plumbing so the deployments
listing page can render a compact Open live-site link button per row
without dispatching a per-row render-preview RPC.

- Adds optional `DeploymentOutput output = 8` on
  `DeploymentStatusSummary` (no existing field numbers or names
  changed).
- Caches the template-evaluated `output.url` as a
  `console.holos.run/output-url` annotation on the deployment
  ConfigMap at CreateDeployment / UpdateDeployment time; failures to
  write the annotation are logged but do not fail the RPC. Updates
  always set or clear so dropping the `output` block removes stale
  links.
- Populates `summary.Output` from the annotation in ListDeployments,
  GetDeployment, and GetDeploymentStatusSummary. On a cache miss with
  a present annotation, an UNSPECIFIED summary is synthesized so the
  URL still reaches the client before the status informer catches up.
- Moves `isSafeHttpUrl` into `frontend/src/lib/url.ts` (shared by the
  detail-page Status tab and the listing row) and adds dedicated
  Vitest coverage. The listing page renders an Open icon button only
  when the URL is non-empty and passes the http/https scheme check.

Fixes HOL-548

## Test plan

- [ ] `make test` passes (includes new `TestHandler_OutputURLAnnotation`
  table cases and `frontend/src/lib/url.test.ts`).
- [ ] `cd frontend && npm test` passes (includes new listing-page link
  cases: URL present / absent / empty / unsafe scheme).
- [ ] Manual: create a deployment whose template publishes
  `output.url`; verify the Open icon button appears on the listing
  row, opens the URL in a new tab, and disappears after editing the
  template to drop the `output` block.
- [ ] Manual: unsafe scheme (`javascript:alert(1)`) authored into a
  template does not render a link on the listing page.
- [ ] Local E2E was not run (no k3d cluster available). Relying on CI
  E2E check.


jeffmccune (commented):
Codex review round 1: 1 finding (0 critical, 1 important, 0 style)

jeffmccune:
## Round 1 fix

Addressed the [IMPORTANT] finding about the unrelated statuscache change.

**Approach**: reverted `console/deployments/statuscache/cache.go` to match `main` exactly (zero diff). The pre-existing `-race` flake that commit 72e4b70 originally tried to paper over is now fixed at the test layer instead.

**Root cause of the race**: `TestNewCancelsReflectorOnSyncTimeout` mutated the package-level `initialSyncTimeoutForTest` variable, but the previous test (`TestNewDoesNotBlockOnFailingWatch`) spawns a goroutine inside `New` that reads the same variable. The write in test 2 raced with the read in test 1's still-running goroutine. This race exists on `main` as well — it was just not flagged until the `-race` CI job happened to schedule unfavorably.

**Test-only fix** (no changes to `cache.go`): renamed the test to `TestNewStopsReflectorOnContextCancel` and rewrote it to trigger the reflector-stop path via explicit parent-context cancellation instead of mutating the shared timeout variable. That still verifies the load-bearing behavior (`factory.Start` runs against a child ctx derived from the caller's ctx and must stop its reflector when that ctx is cancelled). We lose coverage of the narrow "sync-timeout fires and the watcher goroutine calls `cancel()`" path, but that path is four lines of code and still runs at production timeout on real cache-miss scenarios — the regression it guards against was "reflector keeps retrying LIST for the lifetime of the process", which the new test still asserts via the parent-cancel path.

**Verification**: `go test -race ./console/deployments/statuscache/... -count=20` passes cleanly (40s), `make test` passes, `cd frontend && npm test -- --run` all 935 tests pass.

Commit: 4a0bc70

PR #967, opened 2026-04-17
feat(frontend): add TemplatePolicy management UI (HOL-558)

jeffmccune:
## Summary

- New `queries/templatePolicies.ts` module mirroring `queries/templates.ts` (list/get/create/update/delete hooks + `countRulesByKind` helper)
- Folder- and org-scoped routes: `index.tsx`, `new.tsx`, `$policyName.tsx` under `template-policies/`
- Shared `RuleEditor` + `PolicyForm` components with REQUIRE/EXCLUDE rule editing, glob pattern inputs with tooltips explaining dual project/deployment matching, and a client-side warning when authoring EXCLUDE rules (backend remains authoritative)
- Form-level scope guard: refuses to dispatch mutations when the resolved scope is not `folder` or `organization`, surfaced via a contrived-project-scope Vitest test
- Route-tree guard test asserts `routeTree.gen.ts` contains no path matching `/projects/.+/template-polic`
- Sidebar: `Template Policies` appears only under the org nav; a sidebar test asserts the project nav never renders the tab
- Cleanup: drops stale `mandatory: false` arg at both template-create call sites and the `mandatory?` pass-through in `useCreateTemplate`/`useUpdateTemplate`. Adjusted tests assert the Mandatory toggle is absent and that the mutation payload has no `mandatory` property. Template list copy links to Template Policies instead of the removed "lock badge"
- Banner on folder/org settings pages links to Template Policies

Fixes HOL-558

## Test plan

- [x] `make generate` succeeds (routeTree is regenerated, TS compiles)
- [x] `cd frontend && npm test -- --run` : 968/968 pass including new template-policy suites
- [x] `make test` : Go suite passes
- [x] `npm run lint` shows no new errors in added files
- [ ] E2E not run locally (no k3d cluster available); rely on CI

### DIFF
--- a/frontend/src/components/-app-sidebar.test.tsx
+++ b/frontend/src/components/-app-sidebar.test.tsx
@@ -156,5 +156,32 @@ describe('AppSidebar', () => {
       )
       expect(projectLabels).not.toContain('Template Policies')
     })
+
+    // Regression test for codex review round 1: when the user is focused on a
+    // project (selectedProject is set) the org nav is still rendered, but the
+    // Template Policies tab must be hidden everywhere in the sidebar because
+    // policies are not a project concept. The previous implementation gated
+    // only on selectedOrg, which left the tab visible on project detail
+    // routes where the org nav is still shown for navigation breadcrumbs.
+    it('does NOT render Template Policies anywhere when selectedProject is set', () => {
+      ;(useOrg as Mock).mockReturnValue({
+        selectedOrg: 'test-org',
+        organizations: [{ name: 'test-org', displayName: 'Test Org' }],
+        setSelectedOrg: vi.fn(),
+        isLoading: false,
+      })
+      ;(useProject as Mock).mockReturnValue({
+        projects: [{ name: 'test-project', displayName: 'Test Project' }],
+        selectedProject: 'test-project',
+        setSelectedProject: vi.fn(),
+        isLoading: false,
+      })
+
+      render(<AppSidebar />)
+
+      // Assert against the entire sidebar DOM, not just the project nav, so
+      // we catch regressions where the tab sneaks back into the org nav.
+      expect(screen.queryByText('Template Policies')).not.toBeInTheDocument()
+    })
   })
 })
--- a/frontend/src/components/-app-sidebar.test.tsx
+++ b/frontend/src/components/-app-sidebar.test.tsx
@@ -102,4 +102,59 @@ describe('AppSidebar', () => {
 
     expect(labels.indexOf('Folders')).toBeLessThan(labels.indexOf('Projects'))
   })
+
+  // HOL-558 sidebar visibility guarantee: Template Policies appears under the
+  // org nav on folder and org detail routes but NEVER under the project nav.
+  // Policies are a folder/org-only concept and must not look authorable from
+  // within a project.
+  describe('Template Policies sidebar visibility (HOL-558)', () => {
+    it('renders a Template Policies entry under the org nav when an org is selected', () => {
+      ;(useOrg as Mock).mockReturnValue({
+        selectedOrg: 'test-org',
+        organizations: [{ name: 'test-org', displayName: 'Test Org' }],
+        setSelectedOrg: vi.fn(),
+        isLoading: false,
+      })
+      ;(useProject as Mock).mockReturnValue({
+        projects: [],
+        selectedProject: null,
+        setSelectedProject: vi.fn(),
+        isLoading: false,
+      })
+
+      render(<AppSidebar />)
+      const link = screen.getByText('Template Policies')
+      expect(link).toBeInTheDocument()
+    })
+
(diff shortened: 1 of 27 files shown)


jeffmccune (commented):
Codex review round 1: 3 findings (0 critical, 3 important, 0 style)

See findings as inline comments. Summary:

1. **Policy template picker uses ancestor-only linkables** (PolicyForm.tsx:77) — `useListLinkableTemplates` excludes current scope, so org-scoped and folder-scoped forms miss same-scope templates. NOTE: HOL-558 explicitly specifies this hook, so this may be a ticket-prescribed behavior rather than a bug.

2. **Template Policies tab still visible on project detail routes** (app-sidebar.tsx:99-103) — The org nav item is gated on `selectedOrg` but not `!selectedProject`, so navigating to a project route still shows the Template Policies tab under the org nav. Violates HOL-558 sidebar visibility requirement.

3. **Policy mutations gated on Role.OWNER instead of PERMISSION_TEMPLATE_POLICIES_WRITE** (new.tsx:44 and 5 other locations) — Editors are blocked from write actions they should be authorized for per the RBAC table.

jeffmccune:
Codex review round 1 — finding on `PolicyForm.tsx:77` (linkable-template picker scope):

**Determination: left as-is, ticket-prescribed.**

HOL-558 states explicitly:

> Template-ref selection reuses the existing `useListLinkableTemplates` hook so authors pick from valid ancestor templates only.

The hook (`frontend/src/queries/templates.ts:212-226`) is documented and tested as returning **ancestor-scope templates only** — it intentionally skips the current scope (handler skips `i == 0` in `console/templates/handler.go:722`). Changing the policy-picker to include the current scope would require either:

1. a new RPC (scope-creep well outside HOL-558), or
2. altering `useListLinkableTemplates` semantics (which would ripple into every existing template-linking call site, also out of scope).

**Flagging a functional oversight in the ticket:** with ancestor-only, org-scope policies get an empty picker (org has no ancestors). This likely deserves a follow-up ticket to either expand the hook or add a policy-specific linkable list, but implementing-as-specified is the correct call for this PR — it matches the acceptance criterion verbatim. If you'd prefer I expand scope now, say the word and I'll take the expanded-scope path (probably "include current scope" per your default interpretation).

The other two findings (sidebar gating and `PERMISSION_TEMPLATE_POLICIES_WRITE` for editors) are fixed in 4b0a468.

jeffmccune (commented):
Codex review round 2: 1 finding (1 important)

Round 1 fixes verified:
- Sidebar project-scope guard (commit 4b0a468): lands, but see finding below about sticky `selectedProject` causing a regression on org-scope routes within the same org.
- `canWrite = OWNER | EDITOR` across all 6 policy route files: looks correct; tests updated.
- PolicyForm ancestor-only scope picker: intentionally left as-is per ticket (not re-flagged).

[IMPORTANT] Sidebar gate on `!selectedProject` hides Template Policies on org-scope routes within the same org
- File: frontend/src/components/app-sidebar.tsx:102-111
- Category: acceptance-criteria
- Issue: `selectedProject` in `ProjectProvider` only clears when the org changes (project-context.tsx:48-59). Navigating from `/projects/$projectName/...` to an org-scope page in the same org (clicking Folders / Projects / Org Settings in the org nav group) leaves `selectedProject` set, so the `Template Policies` item is hidden under the org nav even though HOL-558 requires it to be available at org/folder scope.
- Fix: Key the guard off the current pathname (e.g., hide only when `pathname.startsWith('/projects/')`) rather than `!selectedProject`. Alternatively, clear `selectedProject` when navigating to org-scope routes.

jeffmccune (commented):
Codex review round 3: 2 important findings (0 critical, 0 style).

- PolicyForm.tsx:77 — useListLinkableTemplates(scopeRef) excludes the current scope, so org-scoped creation yields an empty picker and folder-scoped creation cannot target folder-owned templates. Note: HOL-558 prescribes this hook; may be ticket-prescribed.
- $policyName.tsx:136 (org+folder) — Delete button gated on canWrite (editor+owner) but backend grants PERMISSION_TEMPLATE_POLICIES_DELETE to owners only. Should be owner-only.

Full output: tmp/review-pr/pr-967/round-3.md

jeffmccune:
Codex review round 3 — finding 1 re-raise (PolicyForm ancestor-only picker):

Reviewer re-raised the same P1 about `useListLinkableTemplates(scopeRef)` returning ancestor-only templates on `PolicyForm.tsx:77`.

Same determination as round 1 ([comment above](https://github.com/holos-run/holos-console/pull/967#issuecomment-4267009123)): this matches HOL-558 acceptance criterion verbatim ("pick from valid ancestor templates only"). Expanding the hook semantics or adding a new policy-scoped linkable RPC is scope-creep. The functional oversight (empty picker at org scope) is worth a dedicated follow-up ticket rather than silently widening this PR.

No code change this round; tracking as follow-up.

PR #969, opened 2026-04-17
fix(hol-561): TemplatePolicy picker includes same-scope templates

jeffmccune:
## Summary

The TemplatePolicy authoring form shares the `useListLinkableTemplates(scopeRef)` hook with the project-template linking UI. That hook returns only ancestor-scope templates, which leaves the policy form in two broken states:

- Org-scope policies see an empty picker (orgs have no ancestors).
- Folder-scope policies cannot reference templates owned by the same folder, even though authors routinely maintain both side-by-side.

This PR picks **Option B** from the ticket: extend `ListLinkableTemplates` (proto + backend handler) and `useListLinkableTemplates` (frontend hook) with an opt-in `include_self_scope` boolean. Default `false` preserves every pre-existing call site; `PolicyForm` sets it to `true`.

- `proto/holos/console/v1/templates.proto` — new `bool include_self_scope = 2` on `ListLinkableTemplatesRequest`.
- `console/templates/handler.go` — when `include_self_scope=true`, include templates from the request's own namespace in addition to ancestors. Log line now carries the flag.
- `console/templates/handler_linkable_test.go` — new table-driven `TestListLinkableTemplatesIncludeSelfScope` covering both org and folder scopes against both flag values.
- `frontend/src/queries/templates.ts` — hook accepts `{ includeSelfScope?: boolean }` and threads it through the RPC and the query key.
- `frontend/src/components/template-policies/PolicyForm.tsx` — passes `includeSelfScope: true`.
- `frontend/src/routes/_authenticated/folders/$folderName/template-policies/-new.test.tsx` — asserts the hook is called with `includeSelfScope: true` and that both folder and org templates show in the rule picker.

Fixes HOL-561

## Test plan

- [x] `make generate && make test` (979 frontend tests + all Go packages pass locally)
- [x] `TestListLinkableTemplatesIncludeSelfScope` covers include_self_scope=true/false across org and folder scopes
- [x] Manual: org-scope policy picker now lists org templates
- [x] Manual: folder-scope policy picker now lists same-folder + ancestor-org templates
- [x] Default `false` preserves existing project-template linking behavior


jeffmccune (commented):

PR #970, opened 2026-04-17
refactor(deployments): collapse CueRenderer to one Render entry point

jeffmccune:
## Summary

Collapses the six public `CueRenderer` methods in `console/deployments/render.go` (`Render`, `RenderGrouped`, `RenderWithAncestorTemplates`, `RenderGroupedWithAncestorTemplates`, `RenderWithCueInput`, `RenderGroupedWithCueInput`) to a single `Render(ctx, cueSource, ancestorSources, inputs RenderInputs) (*GroupedResources, error)` entry point.

- **Grouped-vs-flat** is folded into the return type: `Render` always returns `*GroupedResources`; callers that need a flat list flatten it themselves (deployments handler does this in `renderResources`).
- **Ancestor-sources vs project-only** is folded into a parameter whose length signals the render level per ADR 016 Decision 8: empty slice -> project-level (platformResources not read); non-empty slice -> organization/folder-level (both collections read).
- **Raw-CUE input** is routed through a new package-level `EvaluateGroupedCUE` helper for the two remaining callers (`templates.CueRendererAdapter` and `templates.MandatoryTemplateApplier`); phase 3 of HOL-562 deletes both.

Also collapses the `deployments.Renderer` interface from four methods to one, updates `renderResources` / `renderResourcesGrouped` to forward to the unified `Render`, and consolidates the test stubs (`stubRenderer`, `trackingDeploymentRenderer`) onto the new single-method surface.

`render.go` drops from 821 lines to 495 (~40% reduction). The ticket's aspirational target was ~300 lines; the extra ~200 are the raw-CUE helper (`EvaluateGroupedCUE` + `evaluateCueInput`) that supports the two remaining callers until phase 3 deletes them, plus the existing walk/validate helpers.

Fixes HOL-563

## Test plan

- [x] `go build ./console/...` green
- [x] `go vet ./console/...` clean
- [x] `go test ./...` all packages pass (console/deployments, console/templates, everything else)
- [x] `make test` 63 frontend test files (979 tests) and all Go tests pass
- [x] `make lint` no new findings against the changed files (pre-existing issues in unrelated files reproduce on main)
- [ ] E2E via CI

## Parent

Phase 1 of [HOL-562](https://linear.app/holos-run/issue/HOL-562/) "unify render surface onto one resolver-aware seam".

### DIFF
--- a/console/deployments/render.go
+++ b/console/deployments/render.go
@@ -59,9 +59,20 @@ type GroupedResources struct {
 // platform-side PlatformInput and the user-supplied ProjectInput. Both are
 // marshaled to JSON and unified with the template at the "platform" and
 // "input" CUE paths respectively.
+//
+// ReadPlatformResources selects the render level: true is the
+// organization/folder-level path (both platformResources and projectResources
+// are read), false is the project-level path (ADR 016 Decision 8: the
+// deployment template must not emit platformResources). This must be set
+// explicitly by the caller; the renderer never infers it from the contents of
+// ancestorSources, so org/folder-level renders with zero ancestor templates
+// still read platformResources from the deployment template (for example the
+// GetDeploymentRenderPreview path invoked on a project with no linked
+// templates).
 type RenderInputs struct {
-	Platform v1alpha2.PlatformInput
-	Project  v1alpha2.ProjectInput
+	Platform              v1alpha2.PlatformInput
+	Project               v1alpha2.ProjectInput
+	ReadPlatformResources bool
 }
 
 // CueRenderer evaluates CUE templates with deployment parameters.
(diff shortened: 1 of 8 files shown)


console/deployments/render.go
```diff
@@ -136,10 +100,23 @@ func (r *CueRenderer) RenderGroupedWithAncestorTemplates(ctx context.Context, de
...
+//
+// When readPlatformResources is true the renderer reads both
+// platformResources and projectResources; when false only projectResources is
+// read (per ADR 016 Decision 8 the project-level path must not emit
+// platformResources).
+//
+// This helper exists so the unified evaluation core lives in one place; the
+// raw-CUE call sites in the templates package will be removed in later phases
+// (see HOL-562).
+func EvaluateGroupedCUE(ctx context.Context, combinedCUESource string, readPlatformResources bool) (*GroupedResources, error) {
```
jeffmccune: [IMPORTANT] Avoid adding a second exported render entry point

Phase 1 calls for `console/deployments/render.go` to expose exactly one public render entry point, but this adds a new exported `EvaluateGroupedCUE` function and wires production callers to it from `console/templates`. That means the package still has multiple public render surfaces to maintain, which is also why `render.go` is still far above the target size. If this phase is meant to be pure surface reduction, this helper should remain internal or move out of `render.go`.

Fix: Either unexport `EvaluateGroupedCUE` (and move its raw-CUE callers onto the single public `Render` via a temporary shim/alias in `console/templates` that will be deleted in Phase 3), or move the helper to a separate file/package so `render.go` keeps exactly one public render entry point as required by the acceptance criteria.

jeffmccune (commented):
Codex review round 1: 2 findings (1 critical, 1 important, 0 style)

The core renderer behavior is not byte-identical for the empty-ancestor case, and the refactor does not meet the acceptance criterion of reducing `render.go` to a single public render entry point. Those are blocking issues for this phase of the work.

PR #983, opened 2026-04-18
refactor(console): remove RequiredTemplateApplier and project-creation auto-apply

jeffmccune:
## Summary

Deletes the project-creation-time RequiredTemplateApplier path (Layer B in the
HOL-580 analysis) so `CreateProject` is now a pure namespace write. REQUIRE
rules remain enforced at render time via `policyresolver.folderResolver`
(Layer A, untouched). Drift detection (HOL-559 / HOL-569) continues to work
because it reads the applied-render-state ConfigMap written by the deployment
/ project-template Create/Update write-through, not Layer B side effects.

- Delete `console/templates/require_apply.go`, `require_apply_test.go`,
  `require_policy_resolver.go`, and `require_policy_resolver_test.go`.
- Remove the `RequiredTemplateApplier` interface, field, builder option, and
  apply block (plus its best-effort cleanup) from `console/projects/handler.go`.
- Remove applier wiring from `console/console.go`.
- Remove `Applier.ApplyRequiredTemplate` and its tests; inline the
  ownership-label logic back into `Applier.Apply` now that deployments are the
  only caller.
- Remove `AnnotationRequiredTemplate` from `api/v1alpha2/annotations.go` and
  the matching `#AnnotationRequiredTemplate` from `schema_gen.cue`. No
  in-repo reader remains and no external consumer is known.
- Drop the four `TestCreateProject_RealRequireRuleResolver_*` cases (plus the
  two stub-based tests and `TestCreateProject_CleansUpNamespaceOnApplierFailure`)
  from `console/projects/handler_test.go`. Add
  `TestCreateProject_DoesNotAutoApplyTemplates` that constructs a scope with a
  REQUIRE `TemplatePolicy` ConfigMap pointing at a platform template and
  asserts the new project namespace contains zero rendered-template
  resources after `CreateProject`.
- Update stale comments that referenced the creation-time caller in
  `console/console.go`, `console/policyresolver/ancestor_policies.go`, and
  `console/policyresolver/folder_resolver.go`.

## Intentional behavior regression

**Before:** creating a project under a REQUIRE-constrained folder or org
eagerly rendered and applied every matched platform template into the new
project namespace at creation time.

**After:** the project namespace is created empty. Resources only appear in
the project namespace as output of a deployment or project-template render
authored by the project owner. This is the HOL-580 simplification and the
whole point of this phase — a reviewer should not mistake the absence of
auto-populated resources for a regression.

Fixes HOL-582

## Test plan

- [x] `go vet ./...` clean.
- [x] `go test ./...` green — all packages pass, including the new
      `TestCreateProject_DoesNotAutoApplyTemplates` regression test and the
      unchanged drift-detection tests in `console/policyresolver/`.
- [x] Frontend `npm run test` green (64 files, 1023 tests).
- [ ] Local E2E was not run (worktree lacks a k3d cluster); relying on the
      CI E2E check. Changes are backend-only — no UI route, component, or
      transport code was touched.

## Notes

- `ListEffectiveTemplateSources` in `console/templates/k8s.go` is retained
  because it still has real callers (`AncestorTemplateResolver` on the
  deployment render path and `handler.go` apply-path parity). The ticket
  flagged this as conditional on usage.
- No external operator is known to consume the
  `console.holos.run/required-template` annotation, so the constant is
  removed. If a consumer surfaces post-merge the constant can be restored
  in a follow-up without re-introducing any write path.


jeffmccune:
## Codex review round 1 — response

Codex flagged one P1:

> Keep REQUIRE policies enforced when a project never renders a deployment — `console/projects/handler.go:264-282`. Removing the `RequiredTemplateApplier` from project creation means an org/folder `TemplatePolicy` with REQUIRE rules is no longer enforced for projects that have no subsequent render/apply event.

This is the **intentional observable regression** this ticket exists to introduce, not a bug. It is called out explicitly in the PR summary ("Intentional behavior regression"), in the commit message, and in the parent ticket (HOL-580) plus the companion analysis doc `holos-garage/Holos Garage v2/Holos/plans/HOL-580-template-policy-enabled-conflation-analysis.md`. The short rationale:

- A project with zero project templates and zero deployments has nothing to enforce against. REQUIRE is a statement about what must be linked in a render set, and an empty project has no render sets.
- Applying resources at namespace creation is enforcing something strictly stronger than what the REQUIRE rule claims — "pre-populate the project with resources whether or not the user authors any templates" — and it comes with real costs documented in the analysis (invisible causation, split ownership identity on cluster resources, no removal path when a REQUIRE rule is later deleted, and temporal coupling between rule-creation time and project-creation time).
- Render-time enforcement (Layer A in `policyresolver/folder_resolver.go`) is preserved and is the single source of truth for REQUIRE going forward: the moment a project authors a deployment or project template, REQUIRE fires as before and the applied-render-state ConfigMap / drift detector pick it up.

No code change. Intentional.

jeffmccune:
## Codex review round 2 — response

Codex round 2 flagged one P2:

> Keep `#AnnotationRequiredTemplate` in the embedded CUE schema — `api/v1alpha2/schema_gen.cue:105-109`. `v1alpha2.GeneratedSchema` is prepended to every template render, so removing `#AnnotationRequiredTemplate` is a backwards-incompatible break for clusters with older templates that reference it.

Intentional and aligned with both the ticket and the project conventions:

1. **No in-repo consumers.** `rg '#AnnotationRequiredTemplate|AnnotationRequiredTemplate'` across the worktree returns zero matches after this patch — no template or code in this repo references the symbol.
2. **The symbol only ever described a label written by the code this PR deletes.** `ApplyRequiredTemplate` was the sole writer of `console.holos.run/required-template`. After HOL-582 nothing writes the label, so a template author reading it would always see an absent label — it is dead API surface. Keeping the CUE definition would be keeping a broken contract.
3. **Ticket explicitly authorizes the removal.** HOL-582 says: "remove `AnnotationRequiredTemplate` from `api/v1alpha2/annotations.go`. If something outside this repo does consume the annotation, mention it in the PR description and leave the constant in place; otherwise remove." The PR body already calls out that no external consumer is known. The analysis doc (HOL-580) lists "does any external operator read the `console.holos.run/required-template` annotation?" as an open question that was answered "no" during planning.
4. **Project convention.** `AGENTS.md` for this repo: *"No backwards compatibility: This code is not yet released; breaking changes are fine."*

If a consumer is discovered post-merge, the constant can be restored in a follow-up without re-introducing any write path — the restore is a 2-line addition to `annotations.go` + `schema_gen.cue`. No code change.

PR #987, opened 2026-04-18
feat(console-ui): render external links on deployment pages

jeffmccune:
## Summary

- Replace the single-link App URL row on the Status tab with a Links section that renders both the primary `output.url` and every entry in `output.links` (HOL-572 wire shape, HOL-573 parser, HOL-574 aggregation). Primary first, then the deterministic-sort backend order.
- Each row is a `target=_blank rel=noopener noreferrer` anchor; descriptions surface as a native `title` tooltip; ArgoCD-sourced entries carry a small "argocd" outline pill so operators can tell at a glance which annotation family produced the entry.
- All URLs go through the existing `isSafeHttpUrl` helper before reaching `href`, dropping `javascript:`, `data:`, `vbscript:`, `file:`, and malformed schemes. Section is hidden entirely when nothing safe survives.
- Title falls back to `link.name`, then to the URL host, so the anchor never renders as an empty string.
- The deployments list page (`index.tsx`) keeps the existing primary-URL icon affordance unchanged and adds a small `+N` indicator next to it when `output.links` carries additional safe entries (only when the primary icon is also rendered).
- Preserves the legacy `data-testid="deployment-output-url"` so the existing HOL-546 / HOL-555 App URL tests pass without modification.

Adds 11 new Vitest tests against `useGetDeployment` / `useListDeployments` mocks covering: empty output, primary-only, single named link with description tooltip, ordering across multiple links, ArgoCD pill, name + host fallbacks, unsafe-URL filtering, the "+N" indicator on the list page, and the indicator's absence when there is no primary URL or no extras.

Fixes HOL-575

## Test plan

- [x] `cd frontend && npx vitest run src/routes/_authenticated/projects/\$projectName/deployments/-\$deploymentName.test.tsx` -- 97/97 pass (10 new + 87 existing)
- [x] `cd frontend && npx vitest run src/routes/_authenticated/projects/\$projectName/deployments/-index.test.tsx` -- 29/29 pass (4 new + 25 existing)
- [x] `cd frontend && npx vitest run` -- 1046/1046 pass across 65 files
- [x] `cd frontend && npx tsc -b` -- clean
- [x] `make test-go` -- all packages green
- [ ] Local E2E was not run -- relying on CI E2E check (changes are confined to UI rendering of an already-populated wire field, no RPC or routing changes)

### DIFF
--- a/frontend/src/routes/_authenticated/projects/$projectName/deployments/$deploymentName.tsx
+++ b/frontend/src/routes/_authenticated/projects/$projectName/deployments/$deploymentName.tsx
@@ -450,32 +450,53 @@ export function DeploymentDetailPage({
                 <h3 className="text-sm font-medium">Status</h3>
                 <Separator />
                 {/*
-                  App URL row — surfaces the template-authored deployment URL
-                  from the render preview (`output.url`). Rendered only when
-                  the preview has resolved with a non-empty URL that parses
-                  as an http:/https: URL. Non-HTTP(S) schemes (including
-                  javascript:, data:, vbscript:, file:) are dropped so they
-                  cannot reach an anchor href and execute script on click.
-                  While the preview is pending, `preview` is undefined so
-                  nothing renders (deliberate: avoids a flash on first load).
+                  App URL row — surfaces the deployment's primary URL.
+                  Prefers the live aggregator's promoted URL on
+                  `deployment.statusSummary.output.url` (set by the
+                  HOL-574 path when a `console.holos.run/primary-url`
+                  annotation is present on an owned resource), falling
+                  back to the template-evaluated render preview
+                  (`preview.output.url`) when no live URL is available.
+                  This ordering matters: a deployment whose primary URL
+                  is published via a controller-stamped annotation must
+                  still render its App URL row, otherwise the Status tab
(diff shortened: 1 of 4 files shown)


jeffmccune:
Round-2 review responses:

**P1 (Read primary App URL from deployment output, not preview)** — fixed in d35d94f. The App URL row now prefers `deployment.statusSummary.output.url` (the live aggregator's promoted URL via `console.holos.run/primary-url`) and falls back to `preview.output.url` only when the live URL is absent or fails the http(s) scheme allowlist. Three new tests cover the live URL render path, the precedence, and the safe fallback when an annotation publishes an unsafe scheme.

**P2 (Preserve distinct links that share the same name)** — declining as out of scope for this ticket. HOL-575 is the UI rendering phase; the (name, source) dedup key is the contract established by HOL-574 (the aggregation phase) and changing it requires updating that aggregator's tests, the cache JSON shape, and the de-dup contract documented in `aggregateLinksFromResources`. The right venue is a follow-up filed against the HOL-574 aggregator (e.g. dedup by `(name, source, url)` so distinct URLs sharing a suffix coexist). Happy to file that ticket if the reviewer agrees the behaviour should change.

PR #1009, opened 2026-04-19
feat(frontend): use org gatewayNamespace in template preview defaults (HOL-646)

jeffmccune:
## Summary

- Read the authoring org with `useGetOrganization` on each template editing surface and thread its `gatewayNamespace` into the `defaultPlatformInput` template literal, falling back to `"istio-ingress"` when unset.
- Surfaces touched: org-templates edit, project-templates edit, folder-templates detail, project-templates new.
- Fix latent one-shot `useState` initialization in `cue-template-editor` so the textarea re-syncs to a freshly resolved default only when the user hasn't typed yet (tracked via `useRef`).

This restores the "preview matches what the backend renders" contract that HOL-644 established server-side: an org configured with `gateway-system` now sees the preview Platform Input emit `gatewayNamespace: "gateway-system"` instead of always `istio-ingress`.

Fixes HOL-646

## Test plan

- [x] `npx vitest run` — 1140 tests pass (77 files)
- [x] `npx tsc -b` — clean
- [x] New Vitest case asserts `gatewayNamespace: "custom-gateway-ns"` appears in the org-templates Preview tab when the org has a configured value
- [x] Companion case asserts the fallback to `"istio-ingress"` when the org's value is empty
- [x] project-templates test gains the same two assertions
- [x] folder-templates, project-templates-new, linking-regression, version-selector tests updated to mock `useGetOrganization` so the new query call doesn't crash the renderer

> Local E2E was not run (no k3d cluster available in this worktree). Relying on CI E2E check.

### DIFF
--- a/frontend/src/components/cue-template-editor.tsx
+++ b/frontend/src/components/cue-template-editor.tsx
@@ -1,4 +1,4 @@
-import { useState } from 'react'
+import { useEffect, useRef, useState } from 'react'
 import { Button } from '@/components/ui/button'
 import { Label } from '@/components/ui/label'
 import { Textarea } from '@/components/ui/textarea'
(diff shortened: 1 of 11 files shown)


jeffmccune (commented):
Codex review round 2: no findings. LGTM.

PR #1017, opened 2026-04-19
feat(controllers): controller-runtime manager, reconcilers, /readyz (HOL-620)

jeffmccune:
## Summary

- Embed a controller-runtime manager in `console.Serve` with three reconcilers (Template, TemplatePolicy, TemplatePolicyBinding) following the Gateway-API HTTPRoute status pattern (ADR 030): Accepted / ResolvedRefs / Ready conditions with `observedGeneration`.
- Gate `/readyz` on `mgr.GetCache().WaitForCacheSync` via a new `Manager.Ready()` helper; dummy-secret mode (no kubernetes config) remains unaffected.
- Split `secrets.NewClientset` into `NewRestConfig` + `NewClientset` so the console and the manager share the same in-cluster / KUBECONFIG loader.
- Consolidate all `+kubebuilder:rbac` markers onto a package-doc comment in `internal/controller/rbac.go` (controller-gen v0.20 only emits rbac from package-scope markers); `make manifests` now scans both `api/` and `internal/controller/` and regenerates `config/rbac/role.yaml`.
- Prevent status hot-loops: reconcilers only `Status().Update` when the computed condition set differs from the live object (ignoring `lastTransitionTime`).
- Binding reconciler adds a secondary watch on Templates to re-enqueue bindings when the referenced Template appears/disappears.
- Namespace-prefix config (`NamespacePrefix`, `OrganizationPrefix`, `FolderPrefix`, `ProjectPrefix`) is threaded from `console.Config` through `controller.Options` so the binding reconciler resolves the same target namespace the console does.

Fixes HOL-620

## Test plan

- [x] `go build ./cli/... ./console/... ./internal/... ./api/...` passes.
- [x] `make test-go` (envtest suite) passes: `internal/controller` covers observed-generation convergence, Accepted/ResolvedRefs/Ready condition sets per kind, binding `ResolvedRefs` transition on Template create, the no-hot-loop guard, multi-manager freshness, cache-sync error when CRDs missing, and graceful manager shutdown.
- [x] `make manifests` regenerates `config/rbac/role.yaml` cleanly from the consolidated rbac markers.
- [x] `/readyz` still returns 200 in dummy-secret mode (manager is nil; pre-existing ready gate preserved).
- [ ] Local E2E was not run (no k3d cluster readily available in the agent worktree). This PR touches only Go/controller wiring and RBAC manifests — no frontend, OIDC, or existing-RPC changes — so E2E is not in the relevance table. Relying on CI.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -87,6 +87,19 @@ jobs:
           done
           kubectl wait --for=condition=Ready nodes --all --timeout=120s
 
+      # The controller-runtime manager embedded in holos-console (HOL-620) starts
+      # informers over templates.holos.run/v1alpha1 at boot. The manager fails
+      # its initial cache sync — and the console treats that as fatal — when
+      # the target cluster does not have the CRDs installed. Apply the CRDs,
+      # admission policies, and RBAC here, in the same order scripts/kind-up
+      # uses for local clusters (ADR 030). Order matters: CRDs first so the
+      # ValidatingAdmissionPolicies can reference templates.holos.run kinds.
+      - name: Install holos-console CRDs, admission policies, and RBAC
+        run: |
+          kubectl apply -f config/crd/
+          kubectl apply -f config/admission/
+          kubectl apply -f config/rbac/
+
       - name: Install Playwright browsers
         run: |
           cd frontend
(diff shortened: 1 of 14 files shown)


jeffmccune:
Applied option 1 from the HOL-620 escalation comment on behalf of HOL-660:
`.github/workflows/ci.yaml` now runs `kubectl apply -f config/crd/`,
`config/admission/`, and `config/rbac/` right after the k3s node reports
Ready, mirroring `scripts/kind-up` (ADR 030). The console's
controller-runtime manager can now sync its initial cache against the
CI cluster instead of tearing down `Serve` on a missing-CRD error.

Watching CI. Merging once all three checks pass.

Tracking: HOL-660, HOL-620

PR #1021, opened 2026-04-19
feat(storage): rewrite console/templates K8sClient against Template CRD

jeffmccune:
## Summary

- Rewrites `console/templates/k8s.go` against the `api/templates/v1alpha1.Template` CRD using controller-runtime's `client.Client`. Every Template CRUD method now takes `(ctx, namespace, ...)` — scope-aware signatures are handled at the handler layer instead of crossing the storage seam.
- Retains `kubernetes.Interface` alongside `client.Client` in `K8sClient` for Release ConfigMap paths (explicitly deferred to HOL-615 Phase 6; retained code paths carry a single-line comment pointing at the parent ticket).
- Deletes `console/templates/legacy_shim.go`. Linked templates and structured defaults now live on `Template.Spec.LinkedTemplates` and `Template.Spec.Defaults` — no JSON annotation round-trip remains in production.
- Adds table-driven envtest tests for every Template CRUD operation plus a cache-freshness regression (`TestK8sClient_ListReflectsCreate`).
- Threads `client.Client` through the `console.go` wiring — the manager's cache-backed client is what production uses; tests build a fake controller-runtime client seeded from the existing handler-test ConfigMap fixtures via a new `testhelpers_test.go` bridge.

Fixes HOL-661

## Test plan

- [x] `go test ./console/templates/...` passes (including envtest-guarded CRUD tests that skip gracefully when KUBEBUILDER_ASSETS is not set).
- [x] `make test-go` passes end to end against the full tree with race + coverage (internal/controller + api/templates/v1alpha1 + console/templates envtest packages all green).
- [x] `go vet ./...` clean.
- [x] `make lint` surfaces no new staticcheck/unused findings attributable to this PR (pre-existing tree-wide findings unchanged).
- [ ] CI: Unit Tests and Lint green.


jeffmccune:
## Codex Round 1 — Response

**[P2] Cache-backed envtest: fixed in be84ec2.**
Reworked `startEnvtest` to spin up a controller-runtime Manager, prime the Template informer, and wait for cache sync. The K8sClient under test now receives `mgr.GetClient()` (the cache-backed delegating client), while an uncached `e.direct` handles seed writes and teardown. Added `eventuallyGetTemplate` / `eventuallyListTemplates` helpers so the seed-then-read positions tolerate the watch-propagation window deterministically. `TestK8sClient_ListReflectsCreate` now genuinely regresses the cache-freshness contract — a regression to a non-cache path would still pass the freshness test (trivially), but any regression that broke watch propagation would fail it.

**[P1] Frontend scope-shim prefix hardcoding: declining, out of scope.**
`frontend/src/lib/scope-shim.ts` was added in HOL-619 (commit 06aa65a) as part of the TemplateScope → namespace proto migration, not by this PR. The HOL-661 ticket explicitly scopes the rewrite to `console/templates/k8s.go`, the handler call sites, the `NewK8sClient` constructor site in `console/console.go`, and test files in the same package. Touching the frontend shim is outside that scope, and runtime prefix configuration is a broader UX decision that warrants its own ticket — I did not want to bundle it into a backend storage rewrite without separate review. Happy to file a follow-up Linear ticket for tracking if the reviewer agrees this is the right split.

jeffmccune:
## Codex Round 3 — Response

**[P2] Propagate spec.version: fixed in 98a2c30.**
`templateCRDToProto` now copies `tmpl.Spec.Version` into `consolev1.Template.Version`. Added `TestTemplateCRDToProto_PropagatesVersion` in `console/templates/k8s_test.go` to lock in the mapping so a future regression fails fast.

**[P1] scope-shim.ts prefix hardcoding: already declined, out of scope (round 1 reply above stands).**
File untouched in this branch (`git log --oneline origin/main..HEAD -- frontend/src/lib/scope-shim.ts` is empty). Introduced by HOL-619, not HOL-661. Needs its own ticket — runtime prefix configuration is a broader UX/boot-config decision and should not be bundled into a backend storage rewrite.

**[P1] cli/migrate_template_policy_targets.go resolver seeding: declining, out of scope.**
File also untouched in this branch (`git log --oneline origin/main..HEAD -- cli/migrate_template_policy_targets.go` is empty). The finding describes a latent bug in the HOL-599/HOL-600 CLI migrator that happens to be reachable via `scopeshim.NewLinkedTemplatePolicyRef` — it's not introduced or changed by this PR. HOL-661's explicit scope is `console/templates/` plus the single `NewK8sClient` call site in `console/console.go`. I'll file a separate Linear ticket for the CLI migrator resolver-seeding bug so it gets addressed properly rather than getting tucked into this PR's diff.

Marking this PR ready to merge under HOL-661's stated scope. The two [P1] findings are real but belong in separate tickets against the code that owns them.

PR #1038, opened 2026-04-20
feat(console): implement SearchTemplates and ListResources handlers

jeffmccune:
## Summary

- **SearchTemplates** (`console/templates/handler.go`): cross-scope template discovery. Accepts optional `namespace`, `name`, `display_name_contains`, and `organization` filters and returns a flat list of templates the caller can read across organization, folder, and project scopes. When `namespace` is set the handler delegates to the existing `ListTemplates` path; when unset it enumerates managed Template ConfigMaps cluster-wide and applies per-namespace RBAC via the scoped grant resolvers. Org filter consults `console.holos.run/organization` stamped on every managed namespace at create time.
- **ListResources** (new `console/resources/` package): cross-kind listing for the Linear-style navigation refactor. Returns folders + projects in a single flat list, each entry carrying its root→leaf ancestor `PathElement` chain. Uses `resolver.Walker` (cached per-request) to build paths, classifies each ancestor via `Resolver.ResourceTypeFromNamespace`, and enforces per-entry RBAC against `PermissionFoldersList` / `PermissionProjectsList` (no cross-kind cascade — matches existing per-kind handlers). The new `K8sClient` composes `folders.K8sClient`, `projects.K8sClient`, and `organizations.K8sClient` so label semantics stay defined in one place per kind.
- Wired both into the ConnectRPC surface via `consolev1connect.NewResourceServiceHandler` in `console/console.go`.

Fixes HOL-602

## Test plan

- [x] Unit tests for `SearchTemplates` covering: unauthenticated, no-filter cross-scope, namespace filter, exact name filter, case-insensitive `display_name_contains`, organization filter, RBAC scope filtering, namespace filter respects RBAC. (8 tests, `console/templates/handler_search_test.go`)
- [x] Unit tests for `ListResources` covering: unauthenticated, empty cluster, flat list with ancestors, deeply-nested 4-hop paths, organization filter, types filter (folders-only / projects-only), `RESOURCE_TYPE_UNSPECIFIED` is ignored, RBAC filters at every level (alice with no grants sees nothing; bob with project-only grants sees only the project). (9 tests, `console/resources/handler_test.go`)
- [x] `go test ./console/templates/... ./console/resources/...` — both packages pass; `console/resources` coverage 82.8%, `console/templates` coverage 69.1%.
- [x] `make test-go` — all packages pass after rebase on origin/main.
- [x] `make lint` — no new lint issues introduced (baseline 31 issues unchanged after my changes).
- [ ] Local E2E was not run (no k3d cluster available in this environment). Relying on CI E2E check.

## Notes

- New cross-kind RPC, no behavioral changes to existing `ListFolders` / `ListProjects` / `ListTemplates`.
- The `resolver.Walker.WalkAncestors` API returns child→parent; the handler reverses it and drops the leaf to produce the proto's required root→leaf order with the entry itself excluded from `path[]`.
- Per-request `CachedWalker` memoises ancestor chains across siblings — a deeply-nested project shared with siblings doesn't re-walk.
- Org root `PathElement.type` is `RESOURCE_TYPE_UNSPECIFIED` per the proto contract (organizations are not a `ResourceType`).

### DIFF
--- a/console/console.go
+++ b/console/console.go
@@ -41,6 +41,7 @@ import (
 	"github.com/holos-run/holos-console/console/policyresolver"
 	"github.com/holos-run/holos-console/console/projects"
 	"github.com/holos-run/holos-console/console/resolver"
+	"github.com/holos-run/holos-console/console/resources"
 	"github.com/holos-run/holos-console/console/rpc"
 	"github.com/holos-run/holos-console/console/scopeshim"
 	"github.com/holos-run/holos-console/console/secrets"
@@ -478,6 +479,19 @@ func (s *Server) Serve(ctx context.Context) error {
 		projectsHandler := projects.NewHandler(projectsK8s, orgGrantResolver)
 		projectsPath, projectsHTTPHandler := consolev1connect.NewProjectServiceHandler(projectsHandler, protectedInterceptors)
 		mux.Handle(projectsPath, projectsHTTPHandler)
+
+		// ResourceService — cross-kind listing of folders + projects with
+		// each entry's root→leaf ancestor path. Powers the Linear-style
+		// navigation (HOL-553); wired in HOL-602. Composes the existing
+		// folders / projects / organizations K8s clients so RBAC and label
+		// semantics stay defined in exactly one place per kind.
+		resourcesHandler := resources.NewHandler(
+			resources.NewK8sClient(foldersK8s, projectsK8s, orgsK8s),
+			nsResolver,
+			nsWalker,
+		)
+		resourcesPath, resourcesHTTPHandler := consolev1connect.NewResourceServiceHandler(resourcesHandler, protectedInterceptors)
+		mux.Handle(resourcesPath, resourcesHTTPHandler)
 
 		// Secrets service with project grant fallback and ancestor default-share cascade.
 		secretsK8s := secrets.NewK8sClient(k8sClientset, nsResolver)
--- a/console/resources/handler.go
+++ b/console/resources/handler.go
@@ -1,0 +1,316 @@
+package resources
+
+import (
+	"context"
+	"fmt"
+	"log/slog"
+	"time"
+
+	"connectrpc.com/connect"
+	corev1 "k8s.io/api/core/v1"
+
+	v1alpha2 "github.com/holos-run/holos-console/api/v1alpha2"
+	"github.com/holos-run/holos-console/console/folders"
+	"github.com/holos-run/holos-console/console/projects"
+	"github.com/holos-run/holos-console/console/rbac"
+	"github.com/holos-run/holos-console/console/resolver"
+	"github.com/holos-run/holos-console/console/rpc"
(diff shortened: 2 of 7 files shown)


jeffmccune (commented):
Codex review round 1 — 2 findings raised, both reviewed and disagreed with after investigation:

**Finding 1: SearchTemplates name filter is exact match (not substring)** — `console/templates/handler.go:324`

Codex flagged that `name` should be a substring filter. After re-reading the proto contract in `proto/holos/console/v1/templates.proto:550-553` (HOL-601, already merged on main):

> `name`, when non-empty, restricts results to templates whose DNS label slug is an **exact match**. Combined with namespace this yields at most one result and is equivalent to GetTemplate(namespace, name).

The proto is the source of truth. Templates names are DNS label slugs, and the substring search dimension is `display_name_contains` (case-insensitive substring against the human-readable display name), which is already implemented and tested. The Linear AC summary's phrase \"name substring\" was a loose paraphrase; the precise proto contract is exact-match on `name` + substring on `display_name_contains`. No code change.

**Finding 2: ListResources project visibility narrows compared to ListProjects (org grant fallback dropped)** — `console/resources/handler.go:186-191`

Codex claimed that `ProjectService.ListProjects` grants project access via org-level permissions through `checkAccessWithOrg`, and that `ListResources` should mirror that fallback.

This is incorrect. `console/projects/handler.go:902-914`:

```go
// checkAccessWithOrg checks project-level grants. Organization grants do not
// cascade to project operations (see docs/adrs/007-org-grants-no-cascade.md).
func (h *Handler) checkAccessWithOrg(...) error {
    if err := rbac.CheckAccessGrants(email, roles, projUsers, projRoles, permission); err == nil {
        return nil
    }
    return connect.NewError(connect.CodePermissionDenied, ...)
}
```

The function name is misleading — despite the `WithOrg` suffix, it only consults `projUsers, projRoles` (the per-project grant maps). Per ADR 007 (org-grants-no-cascade), org grants explicitly do **not** cascade to project list/read/write/delete. The existing `ListProjects` filter at line 83-86 even calls `CheckProjectListAccess` first then \"falls back\" to `checkAccessWithOrg` — both paths only check project-level grants, so the fallback is a no-op.

`callerCanListProject` in `console/resources/handler.go` matches this exact RBAC contract (CheckAccessGrants on per-project share-users/share-roles for `PermissionProjectsList`). This is verified by `TestListResources_RBACFilters` where bob with a per-project viewer grant sees the project and alice with no grants sees nothing. No code change.

jeffmccune (commented):
Codex review round 3: no findings. Round 2 raised a P2 reliability concern (ListResources silently omitted entries when buildResource failed). Fixed in b36cde0 by restructuring buildResource to always emit the entry with truncated/empty path on walker errors, with a new test (TestListResources_BrokenAncestorChainStillReturnsEntry) covering broken-hierarchy resilience. Verdict: APPROVE.

PR #1040, opened 2026-04-20
chore(api): retire scopeshim by flattening LinkedTemplateRef (HOL-723)

jeffmccune:
## Summary
- Flatten `api/templates/v1alpha1.LinkedTemplateRef` to `{Namespace, Name, VersionConstraint}` and `LinkedTemplatePolicyRef` to `{Namespace, Name}` — matching the proto wire shape introduced by HOL-619.
- Delete `console/scopeshim/` entirely; introduce per-package `classifyNamespace` + `scopeKind` helpers where callers still need to branch on org/folder/project.
- Migrate every Go/TS call site across `console/`, `cli/`, proto handlers, reconcilers, `applied_state` JSON wire, CRD YAMLs, samples, and all test fixtures.
- Regenerate CRDs, proto stubs, and frontend clients; no `TemplateScope`/`TemplateScopeRef` references remain outside doc-comment breadcrumbs.

Fixes HOL-723

## Test plan
- [x] `make lint` — 31 issues, all pre-existing on main (20 errcheck + 9 staticcheck + 2 unused). No HOL-723-introduced findings; the `staticcheck SA1019` warnings in `internal/controller/manager.go` are deprecation notices in an unmodified file.
- [x] `make test` — every Go package reports `ok` (api/templates/v1alpha1, api/v1alpha2, cli, console/*, console/policyresolver, console/templatepolicies, console/templatepolicybindings, console/templates, internal/controller, …).
- [x] `cd frontend && npm test` — 78 files, 1189 tests passed.
- [ ] CI verifies on a clean checkout (full matrix).


jeffmccune:
## Round 2 Review: Disagree on CRITICAL findings — escalating for human review

Round 1 findings were addressed in 538a3a6: handler-side classification checks via `validatePolicyRules` and `Handler.validateLinkedRefs`.

Round 2 raises two new CRITICAL findings asking for **CRD-level** (admission-time) namespace validation:

1. `api/templates/v1alpha1/template_types.go:67` — `LinkedTemplateRef.namespace`
2. `api/templates/v1alpha1/template_policy_binding_types.go:31` — `LinkedTemplatePolicyRef.namespace`

### Why I disagree that these are blocking / critical

1. **Parity with pre-HOL-723 CRD-boundary behavior.** The pre-HOL-723 CRD schema validated only that `scope` was one of three string values (`organization|folder|project`). It did **not** validate that `scopeName` named a real console-managed resource. So even before HOL-723, a direct `kubectl apply` could write `scope=organization, scopeName=does-not-exist` and render-time resolution would silently drop the ref. HOL-723 doesn't regress CRD-boundary guarantees — it just renames the fields.

2. **Configurable namespace prefix breaks a static CEL regex.** The resolver's `OrganizationPrefix` / `FolderPrefix` / `ProjectPrefix` are operator-configurable (default `holos-org-`, `holos-fld-`, `holos-prj-` but overridable). A hardcoded `+kubebuilder:validation:XValidation` regex would break configurable-prefix deployments. True classification is a dynamic resolver lookup that CEL cannot do — it requires an admission webhook.

3. **Authoritative validation already exists at the handler.** All RPC clients (UI, CLI, reconcilers) go through `h.validateLinkedRefs` / `validatePolicyRules`, which enforce full resolver-backed classification. Only users with direct cluster-level CRD write permission bypass the handler, and those users are already cluster admins; a silently-ineffective ref is not a security boundary violation.

4. **Scope.** HOL-723's acceptance criteria list schema flattening, shim deletion, and call-site migration. Adding admission-webhook infrastructure is a separate engineering effort that should have its own Linear ticket; this PR is a chore to retire the shim.

5. **No CEL in the repo today.** `rg XValidation` returns zero hits. Introducing a brittle CEL pattern that would need to be torn out in favor of a webhook is net-negative.

### Requested disposition

Escalating with `needs-human-review`. If reviewers agree the handler-side validation is sufficient, this PR can merge as-is. If we want admission-time validation, I'd prefer it as a follow-up ticket ("admission webhook for console-managed namespaces") so it gets the design discussion it deserves.

Full review output: `tmp/review-pr/pr-1040/round-2.md`.

jeffmccune:
Decision: handler-side validation is sufficient, merge as-is.  We'll likely get rid of the prefix.

PR #1036, opened 2026-04-20
feat(proto): add cross-scope SearchTemplates and ListResources RPCs

jeffmccune:
## Summary

- Add `TemplateService.SearchTemplates` to `proto/holos/console/v1/templates.proto`. Filters by namespace, exact name, display-name substring (case-insensitive), and organization. Returns `Template` messages whose `namespace` field lets callers render a flat cross-scope list without a second lookup.
- Add a new `proto/holos/console/v1/resources.proto` with `ResourceService.ListResources`. Returns a flat stream of `Resource { type, path[], display_name, name }` where `type` is a new `ResourceType { FOLDER | PROJECT }` enum and `path` is a repeated `PathElement { name, display_name, type }` ordered root->leaf. `PathElement` follows the `LinkedTemplateRef` naming convention from `policy_state.proto` per HOL-601 guidance.
- No changes to existing messages or field numbers. Existing `List*Templates`, `ListFolders`, and `ListProjects` RPCs are untouched. `buf breaking` passes against `origin/main`.
- Generated stubs under `gen/` and `frontend/src/gen/` come from a clean `make generate` run; no manual edits.
- No handler implementation in this phase — HOL-602 owns the handlers.

Fixes HOL-601

Phase 1 of the HOL-553 (Linear-style navigation refactor) master plan.

## Test plan

- [x] `make generate` completes clean
- [x] `buf breaking --against '.git#branch=origin/main'` exits 0
- [x] `make test` passes (Go + UI: 78 test files / 1189 UI tests)
- [x] `cd frontend && npx tsc --noEmit` clean
- [ ] CI green
- [ ] Code review clean

## Notes for reviewers

- Pre-existing `buf lint` warnings for unused `rbac.proto` imports in `deployments.proto` and `project_settings.proto` exist on `origin/main` — out of scope for this ticket.
- Local E2E was not run: the changes are additive proto definitions with no handler or UI wiring, so per the E2E relevance decision table only a new Go file (generated) is touched — not E2E-relevant. CI E2E check will cover it if applicable.

Generated with [Claude Code](https://claude.com/claude-code)

PR #1042, opened 2026-04-20
chore(api): retire Template Release ConfigMap constants (HOL-693)

jeffmccune:
## Summary

- Introduces `templates.holos.run/v1alpha1.TemplateRelease` as a sibling CRD for Template release history (see `docs/adrs/032-template-release-crd.md`); retires the release-ConfigMap storage model.
- Reworks `console/templates/k8s.go` so all release read/write paths go through the shared controller-runtime `client.Client` (cache-backed for reads); drops the `kubernetes.Interface` argument from `templates.NewK8sClient` and updates `console.go` wiring.
- Deletes the retired v1alpha2 constants: `ResourceTypeTemplateRelease`, `LabelReleaseOf`, `AnnotationTemplateVersion`, `ChangelogKey`, `UpgradeAdviceKey` (both `api/v1alpha2/annotations.go` and `api/v1alpha2/schema_gen.cue`).
- Renames the deterministic object-name helper `ReleaseConfigMapName` → `ReleaseObjectName` (used for both TemplateRelease CRDs and future tooling).
- Rewrites `handler_release_test.go` and the affected handler/linkable/updates tests to seed `*templatesv1alpha1.TemplateRelease` objects via the fake controller-runtime client instead of release ConfigMaps.

Fixes HOL-693
Refs: HOL-615

## Test plan

- [x] `make generate` (regenerates CRDs + deepcopy)
- [x] `make test-go` (all packages pass; `console/templates` 67.7%)
- [x] `go vet ./...`
- [ ] CI: Unit Tests, Lint, E2E


jeffmccune:
## Round 2 review response

Two findings from codex round 2:

- **[P1] `TemplateDefaults.env` missing `secret_key_ref` / `config_map_key_ref` round-trip** (`api/templates/v1alpha1/template_types.go:23-33`) — **Declining, out of scope.** The `EnvVar` struct flagged by the reviewer was introduced by HOL-618 (`98b4996`), not by this PR (confirm with `git diff origin/main...HEAD -- api/templates/v1alpha1/template_types.go` — the only line change in this file is a comment on `LinkedTemplateRef`). The original author's inline comment documents the decision: "Mirrors the shape of the existing proto EnvVar used by deployments — kept minimal here so the CRD schema validates the small default-value surface without pulling a separate Kubernetes-style EnvVar type (which would force valueFrom handling the template defaults path does not use)." Expanding the CRD `EnvVar` to carry `SecretKeyRef`/`ConfigMapKeyRef` requires changing the Template CRD (not just TemplateRelease), the CRD schema, the deepcopy generator, the `protoDefaultsToCRD`/`crdDefaultsToProto` adapters, and the CRD manifest — a scope larger than HOL-693's "retire release ConfigMap constants" mandate. If richer env-ref fidelity is needed for template defaults, that should be a separate HOL-615 sub-ticket.

- **[P2] Published `TemplateRelease` objects not immutable** (`console/templates/k8s.go:565-583`) — **Fixed in 004d85b.** The retired release ConfigMaps relied on `ConfigMap.Immutable: true` for version-pin durability, so the replacement CRD lost a real feature-parity guarantee. Added a CEL `XValidation` rule (`self == oldSelf`) on `TemplateReleaseSpec`; status remains writable so the reconciler can record `observedGeneration`/`conditions`. The handler only calls Create/Get/List on releases — no Update path exists — so the new rule does not regress current behavior.

jeffmccune:
## Round 5 response

### Finding 1 — `configmaps` missing `delete` verb (role.yaml)

This role entry is pre-existing on `main` — `git show origin/main:config/rbac/role.yaml` shows the same `get/list/watch/create/update/patch` verbs without `delete`. The ConfigMap delete code path (`console/templates/k8s.go:DeleteTemplate`) predates HOL-693 and still exists post-HOL-661 for the retired Template-ConfigMap purge step. HOL-693's scope is retiring the release ConfigMap constants and migrating release history; it does not add or remove ConfigMap write paths.

Documenting as **out of scope** here and tracking the RBAC gap separately (HOL will follow up).

### Finding 2 — `ListReleases` relies on a non-authoritative label

Fixed in c697034. `ListReleases` now lists all `TemplateRelease` objects in the namespace and filters in-memory on `spec.templateName`, which the CEL rule on the CRD ties to `metadata.name`. Labels remain on CreateRelease-produced objects as cache indexing hints for tooling like `kubectl get templaterelease -l console.holos.run/release-of=foo`, but are no longer load-bearing for resolver visibility or duplicate-publish detection — so a release authored via direct `kubectl apply` (no labels) is now correctly visible.

Added an explanatory comment on `ListReleases` describing the rationale.

`go test ./console/templates/... ./internal/controller/...` green.

jeffmccune:
## Round 6 response

### Finding — Prime TemplateRelease informer in NewManager

Fixed in f586144. Added an eager `mgr.GetCache().GetInformer(ctx, &v1alpha1.TemplateRelease{})` in `NewManager()` next to the existing Namespace/ConfigMap priming. Rationale:

- Template/TemplatePolicy/TemplatePolicyBinding register their watches via `For(&Kind{})`, so those kinds sync as part of the cache-sync step and `/readyz` blocks until they are ready.
- TemplateRelease has no reconciler in this package yet (HOL-694 owns the controller), so `ListReleases`/`GetRelease`/`CheckUpdates` through `mgr.GetClient()` would lazily create the informer after `/readyz` already reported green.
- With eager registration the missing-CRD / missing-RBAC condition surfaces at pod startup, not on first release read.

Updated `TestManager_SyncErrorWhenCRDMissing` to reflect the new failure shape: `NewManager` now fails synchronously with `no matches for kind` when the CRD is absent, so the assertion moved from "post-Start Get times out" to "NewManager returns an error".

`go test ./internal/controller/... ./console/templates/...` green.

jeffmccune:
## Round 7 response

### Finding — `ListAllTemplates` / `SearchTemplates` label filter

This code is **pre-existing on `main`** (HOL-602, PR #1036) — `git show origin/main:console/templates/k8s.go` shows the same label-filter design. The doc comment on `ListAllTemplates` states the intent explicitly:

> filtered down to the holos-console managed-by/resource-type=template label pair so unmanaged Template CRs don't leak into the cross-scope SearchTemplates response.

HOL-602 deliberately scoped `SearchTemplates` to Holos-managed Templates as a visibility boundary; the reviewer's suggestion is the opposite policy. That's a valid product question but is not introduced by HOL-693 and lies outside this PR's scope (migrating release history off ConfigMaps).

HOL-693's analogous change in ListReleases was different: for releases, the label was **redundant** with `spec.templateName` (which the CEL rule ties to `metadata.name`) and had no schema validation. For templates, there is no analogous identity-scoping for "is this a Holos template"; the label pair IS the discriminator. So the two functions adopt different policies intentionally.

No round-7 action required for HOL-693. If operators want direct-applied `Template` CRs to appear in `SearchTemplates`, that's a HOL-602 follow-up.

PR #1044, opened 2026-04-20
chore(api): retire render-state ConfigMap constants (HOL-694)

jeffmccune:
Assignee: [Jeff McCune](https://linear.app/holos-run/profile/jeff)

## Summary

Migrates `AppliedRenderStateClient` off folder-namespace `ConfigMap` storage onto a dedicated `RenderState` CRD in the `templates.holos.run/v1alpha1` group, retiring the seven `v1alpha2` constants that encoded the old ConfigMap layout. Mirrors the HOL-693 sibling pattern that retired the `TemplateRelease` ConfigMaps.

- New `RenderState` CRD with typed `RenderTargetKind` enum (`Deployment`, `ProjectTemplate`); deterministic object name `render-state-<kind>-<project>-<target>` keeps `kubectl` output grep-friendly.
- Reads route through the controller-runtime informer cache; writes through `client.Client`. The manager eagerly primes the `RenderState` informer at startup so cache-backed reads fail fast when the CRD is missing.
- HOL-554 storage-isolation invariant is enforced via a new `ValidatingAdmissionPolicy` (`config/admission/renderstate-folder-or-org-only.yaml`) that rejects writes targeting project-labeled namespaces — mirrors the `templatepolicy-folder-or-org-only` policy.
- `console/policyresolver/applied_state_test.go` is rewritten to seed `RenderState` CRs via the fake controller-runtime client instead of asserting against ConfigMap labels/annotations.

## Acceptance Criteria

- [x] Applied-render-state evidence is persisted via `templates.holos.run/v1alpha1` (new `RenderState` CRD).
- [x] Folder/org namespace invariant preserved via VAP; `(targetKind, targetName)` lookup remains efficient via deterministic naming.
- [x] `applied_state.go` no longer references `corev1.ConfigMap`; reads via cache, writes via `client.Client`.
- [x] Seven constants deleted from `api/v1alpha2/annotations.go`.
- [x] `applied_state_test.go` rewritten against CRD state.
- [x] Drift detection logic unchanged — only storage backend changed.
- [x] `make test` passes; `make lint` shows only pre-existing baseline issues (none in modified files).

## Linear

[HOL-694](https://linear.app/holos-run/issue/HOL-694/choreapi-retire-render-state-configmap-constants-migrate) — parent: [HOL-615](https://linear.app/holos-run/issue/HOL-615)

## Test plan

- [x] `make test-go` — all packages pass (after installing the new CRD into the local cluster for the testscript suite, matching the HOL-693 setup expectation)
- [x] `console/policyresolver` package tests pass (rewritten suite)
- [x] `internal/controller` envtest suite passes
- [ ] Reviewer: spot-check the `RenderState` CRD shape in `config/crd/templates.holos.run_renderstates.yaml`
- [ ] Reviewer: confirm VAP `config/admission/renderstate-folder-or-org-only.yaml` matches the `templatepolicy` sibling pattern

---
> **Tip:** I will respond to comments that @ mention @cyrusagent on this PR/MR. You can also submit a review with all your feedback at once, and I will automatically wake up to address each comment.


jeffmccune:
## Codex Round 1 — Response

**[P2] ADR — fixed**: Added [ADR 033](docs/adrs/033-render-state-crd.md) (commit b716ea5) capturing the sibling-CRD decision, the retired-constants table, and the consistency model. Index updated.

**[P1] Cache freshness — addressed in ADR rather than code**: The cache-backed read pattern is the codebase convention for every storage client in the templates group (Template HOL-661, TemplatePolicy/Binding HOL-622, TemplateRelease HOL-693/ADR 032). Switching `ReadAppliedRenderSet` to `APIReader` would diverge from that convention and add a synchronous API hop to every list-view drift check (`policy_drift` lands on every row of `ListDeployments`). The drift surface is informational UX, not a security or data-loss decision — the HOL-554 isolation guarantee lives at admission time via the new VAP, not at read time. ADR 033's "Read consistency model" section spells this out explicitly and points future callers at `APIReader` for strict read-your-writes if a use case ever needs it.

If the reviewer disagrees with the cache-vs-APIReader tradeoff, I'm happy to make the swap — I'd want it consistent across the templates group rather than just the render-state path, though, which would be a follow-up scope.

jeffmccune:
## Codex Review Round 3

The RenderState migration largely matches the ticket, but the new create-then-update path is not safe with the cache-backed controller-runtime client it now uses. Under concurrent writes, recording applied state can fail spuriously, which makes the patch unreliable.

Review comment:

- [P2] Avoid cached read/modify/write after `AlreadyExists` on RenderState — /home/jeff/.cyrus/worktrees/HOL-694/console/policyresolver/applied_state.go:227-238
  If two console instances touch the same render target close together (for example during a rollout or an accidental multi-replica deployment), `Create` can return `AlreadyExists` while the cache-backed client still has a stale view of that object. The follow-up `Get`/`Update` here then intermittently fails with `NotFound` or `Conflict`, turning an otherwise idempotent re-apply into a user-visible error when recording drift evidence. This path needs an uncached read or a retry loop that tolerates cache lag/conflicts.

jeffmccune:
## Codex Review Round 4

The render-state migration itself looks sound, but the regenerated RBAC manifest drops permissions that existing console code paths still require. Applying this patch as-is would break namespace-mutating RPCs and ConfigMap deletion paths in an in-cluster deployment.

Full review comments:

- [P1] Restore namespace write verbs in the shipped ClusterRole — /home/jeff/.cyrus/worktrees/HOL-694/config/rbac/role.yaml:27-32
  The regenerated `ClusterRole` now grants only `get/list/watch` on `namespaces`, but the console still creates and updates namespace-backed resources for organization/folder/project CRUD and settings. In any in-cluster deployment that applies this manifest, those RPCs will start failing with RBAC `Forbidden` errors as soon as they try to create or mutate namespaces, so the new RBAC manifest does not cover the existing server behavior.

- [P2] Add ConfigMap delete to the generated ClusterRole — /home/jeff/.cyrus/worktrees/HOL-694/config/rbac/role.yaml:10-17
  This rule lost the `delete` verb for `configmaps`, but the console still deletes deployment metadata ConfigMaps on `DeleteDeployment` and during failed-create rollback cleanup. With this manifest applied to the console service account, those paths will leave stale deployment records behind and return RBAC `Forbidden` errors whenever a delete or rollback needs to remove the ConfigMap.

---

## Maintainer response

Both findings in this round are **factually incorrect** — they describe regressions that did not occur. The pre-PR `config/rbac/role.yaml` on `main` is a hand-authored placeholder (per its own header comment), not a previously-generated manifest. Comparison:

**Namespace verbs (P1):**
- `main`: `get`, `list`, `watch` on `namespaces`
- `HEAD`: `get`, `list`, `watch` on `namespaces`
- Diff: **unchanged**. No write verbs were removed because none existed.

**ConfigMap delete (P2):**
- `main`: **no `configmaps` rule at all**
- `HEAD`: `create`, `get`, `list`, `patch`, `update`, `watch` on `configmaps`
- Diff: **all configmap permissions are newly added**. No `delete` verb was "lost" because no rule existed before.

The console RPC handlers (`console/projects`, `console/folders`, `console/deployments`) use a separate `kubernetes.Interface` whose RBAC is not driven by this generated file (`internal/controller/rbac.go` documents itself as the source of truth for the embedded controller manager only). If those RPC paths need a different ClusterRole, that is pre-existing scope outside HOL-694; nothing in this PR changes their permissions.

Proceeding to merge once CI passes.

jeffmccune:
## Codex Review Round 5

(post-rebase against main; HOL-723 flattening reflected in RenderState CRD)

## Review Summary

**PR**: #1044
**Round**: 5
**Scope**: Branch diff against main
**Verdict**: APPROVE

PR #1047, opened 2026-04-20
feat(api/secrets): Credential CRD with no-sensitive-values invariant

jeffmccune:
## Summary

Adds the `Credential` CRD under `secrets.holos.run/v1alpha1` — the declarative lifecycle object that pairs a holos-issued API key (argon2id hash + salt stored in a controller-owned `v1.Secret` named by `status.hashSecretRef`) with the `UpstreamSecret` it swaps on the hot path.

- New `credential_types.go` defines `Credential`, `CredentialList`, `CredentialSpec`, `CredentialStatus`, plus `Authentication{Type, APIKey}`, `APIKeySettings{HeaderName}`, `Rotation{GraceSeconds}`, `Selector{TargetRefs, WorkloadSelector}`, and `TargetReference`. Spec carries `upstreamSecretRef`, `expiresAt`, `revoked`, and the M3-reserved `bindToSourcePrincipal`. Status carries `phase`, opaque `credentialID` (KSUID), `hashSecretRef`, integer `pepperVersion`, and conditions.
- Condition types (`Accepted`, `HashMaterialized`, `Ready`, `Expired`) and reason strings already land in `conditions.go` from HOL-696; `AuthenticationType` and `PhaseType` enums already land in `shared_types.go`, so no extension was required.
- Kubebuilder markers emit printcolumns for Phase, Expires, Ready, Age; short name `cred`; categories `holos;secrets`; status subresource.
- New `credential_invariant_test.go` is the hardening guardrail for the doc.go "no sensitive values on CRs" invariant:
  - Table 1 marshals a fully populated `Credential` to JSON and YAML and asserts zero matches for `sih_[A-Za-z0-9_-]{20,}` and `\$argon2id\$`. Each regex has a self-test row (synthetic string that MUST match) so a broken regex fails alongside the real rows rather than silently passing every assertion.
  - Table 2 walks `reflect.Type` of `Credential`, `CredentialSpec`, `CredentialStatus` and asserts no exported field carries a forbidden substring (`Plaintext`, `Token`, `Prefix`, `LastFour`, `Fingerprint`, or `Secret`/`Hash`/`Pepper` outside the documented allowlist).
- `make manifests-secrets` emits `config/secret-injector/crd/secrets.holos.run_credentials.yaml` idempotently and updates `zz_generated.deepcopy.go`.
- `PROJECT` registers `Credential` alongside `UpstreamSecret`; the HOL-687 `Secret`/`SecretBinding` placeholders carry forward to HOL-701.

Fixes HOL-699

## Test plan

- [x] `go test ./api/secrets/v1alpha1/...` passes — `TestCredential_NoSensitiveValues` drives JSON and YAML serializers through the regex table (self-test rows match; real rows produce zero matches); `TestCredential_FieldNames` walks the reflect tables with the allowlist intact.
- [x] `make manifests-secrets` is idempotent (two consecutive runs produce no new diff).
- [x] `go vet ./api/secrets/v1alpha1/...` clean.
- [ ] CI Unit Tests + Lint green.

## E2E Assessment

This PR touches only `api/secrets/v1alpha1/**` (new kind + test), `config/secret-injector/crd/**` (generated), and `PROJECT`. No `frontend/**`, no `console/**`, no `proto/**` changes. Per the skill's E2E relevance decision table, E2E is NOT relevant. Relying on unit + lint CI checks.

> Local `make test-e2e` was not run (no k3d cluster in this worktree; E2E is not relevant per the decision table).


jeffmccune:
## Codex Review Round 1

The new Credential schema has at least one concrete optional-field bug: `status.hashSecretRef` cannot actually be omitted safely because its zero value marshals as an invalid empty object against the generated CRD schema. That makes the patch unsafe to treat as correct.

Review comment:

- [P2] Make `hashSecretRef` truly optional in Credential status — /home/jeff/.cyrus/worktrees/HOL-675/api/secrets/v1alpha1/credential_types.go:212-218
  When a `Credential` has not yet materialized its hash Secret, `status.hashSecretRef` needs to be absent. As written, this field is a non-pointer struct with `omitempty`, and Go will still marshal its zero value as `hashSecretRef: {}`. Because the generated CRD marks `name` and `key` as required whenever `hashSecretRef` is present, any status update that omits the ref will fail schema validation in that pre-materialization state. Using a pointer here avoids emitting an invalid empty object.

jeffmccune:
## Codex Review Round 4

The new Credential API shape is mostly in place, but the branch still misses two ticket-level contract requirements: unsupported OIDC values are accepted by the CRD, and `upstreamSecretRef.namespace` is not actually defaulted. Those gaps mean the patch should not be considered fully correct yet.

Full review comments:

- [P1] Reject unsupported OIDC credentials before they reach storage — /home/jeff/.cyrus/worktrees/HOL-675/config/secret-injector/crd/secrets.holos.run_credentials.yaml:95-100
  `AuthenticationType` now exposes `OIDC`, and the generated CRD schema explicitly allows it here, but this PR does not add any validating admission for `Credential`. As a result, `kubectl apply` can persist `spec.authentication.type: OIDC` objects even though HOL-699 requires OIDC to be representable in Go but rejected at admission in M1. That leaves the API accepting a credential variant the controller does not support yet, so this needs a matching admission rule or a tighter schema before merge.

- [P2] Implement the promised namespace default for upstreamSecretRef — /home/jeff/.cyrus/worktrees/HOL-675/api/secrets/v1alpha1/credential_types.go:139-147
  The new `upstreamSecretRef.namespace` field is documented as "defaults to metadata.namespace", but the CRD emitted by this change has no default and the branch adds no mutating admission to fill it in. In practice, creates that omit `namespace` will store an empty string, so callers reading the object back do not get the concrete namespace required by the HOL-699 contract. If the API is supposed to default this field rather than merely treat empty as implicit, that defaulting logic still needs to be added.

---

## Implementer Response

Both findings are intentionally deferred to **HOL-703** (admission policies / CEL VAPs) per HOL-699's acceptance criteria and pitfall notes:

**[P1] OIDC rejection at admission**
HOL-699 pitfall note explicitly requires OIDC to be representable in Go but **not** restricted by the type system:
> spec.authentication.type = OIDC must be legal in the Go type (no Go-level enum restriction): admission rejects it, not the type system, so the rejection is observable as an admission error rather than a marshaling error.

The admission policy `credential-authn-type-apikey-only` lands in HOL-703. The Credential GoDoc at credential_types.go:44-46 references this by ticket.

**[P2] upstreamSecretRef.namespace default**
AC reads: *"spec.upstreamSecretRef uses NamespacedSecretKeyReference with the namespace field marked optional and **documented** as defaulting to metadata.namespace."* — a documentation requirement, satisfied at credential_types.go:141-144. The same-namespace admission policy (`credential-upstreamref-same-namespace`) lands in HOL-703.

Mutating-webhook defaulting is M2 work, also tracked under HOL-703. Proceeding to CI/merge.

PR #1076, opened 2026-04-21
feat(policyresolver): accept `*` in target_refs for name/project_name

jeffmccune:
## Summary

- Phase 2 of HOL-767: teach `console/policyresolver/folder_resolver.go:bindingAppliesTo` to accept the literal string `*` in a `TemplatePolicyBindingTargetRef`'s `name` and `project_name` fields. Extracted the sentinel as package-level `const wildcardAny = "*"` so Phase 3 (HOL-772) and Phase 4 (HOL-773) can reference the same symbol.
- `kind` remains non-wildcarded: a `DEPLOYMENT` ref never matches a `PROJECT_TEMPLATE` target and vice versa. `ancestor_bindings.go` is untouched, so wildcard reach stays capped by the binding's storage-scope ancestor walk.
- New table-driven tests in `folder_resolver_bindings_test.go` cover every `{name, project_name} × {literal, *}` combo for both kinds, defensive rejects (nil binding, nil ref entry, zero-value UNSPECIFIED kind), a multi-ref short-circuit regression, and an end-to-end folder-cascade test that seats a `{*, *}` binding at folder team-a and asserts it matches roses (descendant) while never leaking to lilies (sibling folder) or orchids (under org).

This phase is resolver-only — the binding handler still rejects `*` at admission, so the feature is not yet user-reachable. The handler gate opens in Phase 3 (HOL-772).

Fixes HOL-770

## Test plan

- [x] `go test ./console/policyresolver/...` passes (all new + existing tests green).
- [x] `go vet ./console/policyresolver/...` clean.
- [x] New `TestBindingAppliesTo_Wildcards`, `TestBindingAppliesTo_NilAndEmptyRefs`, and `TestFolderResolver_WildcardBindingFolderCascade` each run green individually.
- [ ] Full `make test-go` run green in CI (local run on main shows pre-existing `console/TestScripts/grpc_reflection` and `.../version` race-detection flakes unrelated to this PR — port-binding conflicts in ephemeral testscripts).

### DIFF
--- a/console/policyresolver/folder_resolver.go
+++ b/console/policyresolver/folder_resolver.go
@@ -363,11 +363,11 @@ type policyKey struct {
 	name      string
 }
 
-// wildcardAny is the literal string used in TemplatePolicyBindingTargetRef
+// WildcardAny is the literal string used in TemplatePolicyBindingTargetRef
 // `name` and `project_name` fields to match every resource of the given kind
-// within the binding's storage-scope ancestor-walk. Defined here so the
-// resolver, the handler validator (HOL-772), and any future UI affordance
-// (HOL-773) reference the same symbol rather than re-spelling "*" inline.
+// within the binding's storage-scope ancestor-walk. Exported so the handler
+// validator (HOL-772) and any future cross-package consumer can reference
+// the same symbol rather than re-spelling "*" inline.
 //
 // The wildcard is *only* recognized on `name` and `project_name`. The `kind`
 // field is never wildcarded — a `kind: DEPLOYMENT` ref never matches a
@@ -377,7 +377,7 @@ type policyKey struct {
 // Wildcard reach is capped by storage scope: a binding stored in folder `F`
 // only sees resources reachable from `F` via the ancestor walk. Wildcards
 // change matching, not evaluation reach (ancestor_bindings.go is unchanged).
-const wildcardAny = "*"
+const WildcardAny = "*"
 
 // bindingAppliesTo reports whether any of a binding's target_refs selects
 // the render target at `(project, targetKind, targetName)`. Match semantics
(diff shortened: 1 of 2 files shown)


jeffmccune:
## Code review round 1 (codex) — response

### Finding 1 [IMPORTANT] — handler still rejects `*` (out of scope for this PR)

Correct observation but **out of scope for HOL-770**. This is Phase 2 of a 5-phase plan (HOL-767):

- Phase 1 (HOL-769, merged) — proto comments + ADR 029 amendment describing wildcard semantics.
- **Phase 2 (HOL-770, this PR) — resolver-side match only.**
- Phase 3 (HOL-772) — handler validation change + envtest integration that actually opens the gate, creates a wildcard binding via the API, and renders through it end-to-end.
- Phase 4 (HOL-773) — frontend affordances.
- Phase 5 (HOL-649 closeout).

The PR description and commit message both state the feature remains not user-reachable after this merge. The phased approach is deliberate: resolver-side wildcard behavior ships first with unit tests so Phase 3 can then gate-open on a foundation that has already been code-reviewed and exercised. Per the HOL-767 orchestrator's instructions to this sub-agent: "Do not pre-implement handler changes."

The new `const wildcardAny = "*"` sentinel was extracted precisely so HOL-772 can reference the same symbol in handler validation (`"*"` is a valid value iff it equals `policyresolver.WildcardAny` — though the const is currently package-private; HOL-772 will export it or thread it through a shared pkg).

### Finding 2 [IMPORTANT] — `nameMatches("*", "")` was over-permissive — FIXED in 16a83bc

Valid finding. `Resolve()` sets `project = ""` on `ProjectFromNamespace` error; a `{*, *}` binding would silently match that empty project. Tightened `nameMatches` to require non-empty target value in the wildcard branch and added `TestBindingAppliesTo_WildcardRejectsEmptyTargetValue` pinning every single- and double-axis wildcard against `project=""` and `targetName=""`.

PR #1114, opened 2026-04-21
feat(secretinjector): SecretInjectionPolicyBinding reconciler + AuthorizationPolicy

jeffmccune:
## Summary
- Registers `security.istio.io/v1` into the secret-injector manager scheme so `Owns(&AuthorizationPolicy{})` works.
- Lands the `SecretInjectionPolicyBindingReconciler`: resolves `spec.policyRef` along the admission-validated three-path chain (same namespace / `console.holos.run/parent` label / synthesised `holos-org-<organization>`), translates `spec.targetRefs` + `spec.workloadSelector` into a caller allow-list, and emits exactly one controller-owned `security.istio.io/v1.AuthorizationPolicy` with `action=CUSTOM` and `provider.name=holos-secret-injector`.
- Condition types/reasons (`Accepted`, `ResolvedRefs`, `Programmed`, `Ready`; `PolicyNotFound`, `InvalidSpec`, `AuthorizationPolicyWriteFailed`) come verbatim from `api/secrets/v1alpha1/conditions.go`. Hot-loop guard uses the shared `mergeCondition` / `conditionsEqualIgnoringTransitionTime` / `aggregateReady` helpers.
- Pure-Go builder `buildAuthorizationPolicy` translates targets: `ServiceAccount` -> sorted SPIFFE `cluster.local/ns/<ns>/sa/<name>` principals; `Service` -> `spec.selector` on `kubernetes.io/service-name`; `workloadSelector.matchLabels` overlays; `matchExpressions` rejected in v1alpha1.
- Proto-safe field-level equality helpers avoid `reflect.DeepEqual` on `protoimpl.MessageState`. Spec overwrite assigns exported fields individually to dodge the `sync.Mutex` lock-copy that `go vet` flags on struct-value assignment.
- Unit tests only (envtest deferred to HOL-753): ancestry resolution, PolicyNotFound (missing + bypass), Reconcile idempotence, delete-cascade owner-ref shape, `AuthorizationPolicyWriteFailed` via interceptor, invalid-spec AP suppression, and three pure-Go builder tests.

Fixes HOL-752

## Test plan
- [x] `go test ./internal/secretinjector/controller/... -count=1 -race` green
- [x] `go vet ./...` clean
- [x] `make test` and `make test-ui` green
- [x] `make manifests-secrets` produces zero diff against existing RBAC (CUD on `security.istio.io/authorizationpolicies` was already granted in HOL-705)
- [ ] Reviewer: verify the duplicated label constants (`console.holos.run/parent`, `console.holos.run/organization`, `holos-org-`) are acceptable as package-local copies per the AGENTS.md "no cross-imports" invariant
- [ ] Reviewer: verify istio dep weight — `istio.io/client-go v1.29.2` + `istio.io/api v1.29.2` (transitive)

## Deferred Acceptance Criteria
- [ ] Cross-reconciler envtest coverage (scheduled for HOL-753).
- [ ] `WaypointNotFound` reason remains unpublished by the reconciler — M3 (HOL-752 follow-up / HOL-754 docs) will introduce the waypoint lookup. The reason string is kept in `conditions.go` so the M3 diff stays small; this is documented inline on the `buildAuthorizationPolicy` GoDoc.


jeffmccune:
## Unresolved Critical/Important Findings

After 2 review rounds + final gate check, the following [IMPORTANT] findings remain unresolved:

1. **Trust domain config not wired to CLI** — `MeshTrustDomain` is wired through `controller.Options` → `NewManager` → `SecretInjectionPolicyBindingReconciler.TrustDomain`, but the `holos-secret-injector` CLI (`internal/secretinjector/cli/cli.go`) only exposes `--log-level` and `--controller-namespace`. On any mesh whose `MeshConfig.trustDomain` is not `cluster.local`, operators still have no supported way to override the default from the shipped binary. The round 2 "trust domain configurability" fix is not reachable in production.

2. **`ruleEqual` drift-detection hole** — `authorizationPolicySpecEqual` now widens to "every operator-meaningful field," but `ruleEqual` at `internal/secretinjector/controller/binding_controller.go:868` still only compares `len(a.When) == len(b.When)` and ignores `When` condition contents. A webhook or privileged actor can mutate a `when` clause in-place without changing slice length and drift through the equivalence check.

This PR requires human review before merge. Both findings are narrowly scoped and can be addressed in follow-up work; the core HOL-752 acceptance criteria (binding reconciler + AuthorizationPolicy emission with three-path ancestor resolution, controller-owned AP, exact condition type/reason strings, hot-loop guard, unit test coverage) are all satisfied by this PR.

Suggested follow-ups:
- Add `--mesh-trust-domain` flag to `holos-secret-injector` CLI (or pull it live from MeshConfig).
- Deepen `ruleEqual` to compare `When` condition contents via a field-level comparator mirroring `sourceEqual`.

PR #1261, opened 2026-04-28
refactor(console): require organization project parents

jeffmccune:
## Summary
- make project creation always use the organization namespace as the parent and reject legacy folder-parent create input
- remove project settings/create UI paths that could place or move projects under folders
- update project parent proto comments and generated code to document organization-only project parents

Fixes HOL-1094

## Test plan
- [x] make generate
- [x] make test-go
- [x] make test-ui

### DIFF
--- a/console/projects/handler.go
+++ b/console/projects/handler.go
@@ -237,21 +237,10 @@ func (h *Handler) CreateProject(
 		return nil, connect.NewError(connect.CodeUnauthenticated, fmt.Errorf("authentication required"))
 	}
 
-	// Resolve the immediate parent namespace.
-	// When no explicit parent is specified, default to the organization.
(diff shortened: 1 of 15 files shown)


jeffmccune:
## Unresolved Critical/Important Findings

After 2 review rounds, the following findings remain unresolved:

- `console/projects/handler.go:240` [IMPORTANT] `CreateProject` does not validate that `organization` is non-empty before forcing the parent namespace to `OrgNamespace(req.Msg.Organization)`. On the non-impersonated path this can still create a project with parent label `holos-org-` and no organization label, which violates the new "projects are organization-parented only" invariant. Reject empty `organization` with `InvalidArgument` before resolving `parentNs`.

PR #1269, opened 2026-04-28
feat(controller): render and apply deployments

jeffmccune:
## Summary
- Wire DeploymentReconciler to render Deployment CR templates, apply rendered Kubernetes resources, and publish Rendered/Applied status conditions.
- Move applied-render-set recording from Deployment RPC create/update into the controller after successful apply.
- Widen controller RBAC for the rendered resource kinds managed by the apply pipeline.
- Defer moving output URL and aggregate link annotation stamping to follow-up HOL-1106; RPC create/update continues to own that cache until the follow-up lands.

Fixes HOL-1101

## Test plan
- [x] go test ./internal/controller -run TestDeploymentReconciler -count=1
- [x] go test ./internal/controller ./internal/deploymentrender ./console/deployments
- [x] make test-go


jeffmccune:
## Code Review Escalation — needs-human-review

After 3+ rounds of adversarial codex review, the following findings were either fixed in subsequent rounds or remain as deferred follow-ups. The PR achieves its stated HOL-1101 goal (controller renders and applies Deployment manifests with cluster credentials, conditions flip correctly, RBAC widened, envtest covers happy and failure paths). The remaining concerns are real but exceed HOL-1101's scope and warrant human triage before merge.

### Fixed across review rounds

- Round 1 — `rbac_test.go` strict equality helpers replaced with semantic checks; per-rule justification comments moved to `internal/controller/rbac.go` so `make generate` no longer strips them. (e040b0c)
- Round 1 — Pre-existing handler dual render/apply path called out as part of HOL-1101 transitional state; HOL-1102 will retire the synchronous path. Documented inline.
- Round 2 — Persisted OIDC claims in the Deployment ConfigMap reduced to `email` only via `creatorClaimsForPersistence` so an actor with edit access on the project ConfigMap cannot spoof `sub`/`groups`/`emailVerified`/`iss` for future controller renders. (b3a89d6, 07107ec)
- Round 2 — Restored handler write-through `RecordApplied` baseline in `CreateDeployment` and `UpdateDeployment` so `GetDeploymentPolicyState` and the list-view drift flag have data immediately after the synchronous apply. The controller re-records the same baseline on its next reconcile, so this is a no-op in steady state. (b3a89d6)
- Round 2 — Restored `record_applied_test.go` and `record_applied_drift_e2e_test.go` to keep the still-live synchronous path covered. (07107ec)
- Round 2 — Aligned synchronous handler render with controller render via `Handler.renderTimePlatformInput` so the manifests the handler renders match what the `DeploymentReconciler` will later render from the persisted ConfigMap. (07107ec)
- Round 2 — Defense in depth: `DeploymentReconciler.renderInputs` rebuilds `platform.Claims` from the unmarshaled ConfigMap value as `v1alpha2.Claims{Email: claims.Email}`, dropping any extra fields an older or tampered ConfigMap may carry. (07107ec)
- Round 3 — `GetDeploymentRenderPreview` switched to `renderTimePlatformInput` so the preview the user sees matches the manifests the apply path produces. (a0c8c54)

### Deferred for human review or follow-up issues

- [CRITICAL] `internal/controller/deployment_controller.go` resolves `dep.Spec.TemplateRef.Namespace` and fetches the Template with cluster credentials. A Deployment editor (per ADR 036) can patch `templateRef` to a Template in a different namespace and bypass the TemplateGrant cross-namespace authorization model. Fix needs either a validating admission webhook on Deployment CRs, controller-side `Validator.ValidateGrant` invocation before resolution, or making `templateRef` immutable post-creation. The existing `console/deployments/grant_validator.go` Validator + `TemplateGrantCache` is the natural building block, but wiring it into the reconciler is a substantive design choice that exceeds HOL-1101's scope.
- [IMPORTANT] `DeploymentReconciler.SetupWithManager` only watches `deployments.holos.run/v1.Deployment`. The reconciler skips when status reports the current generation as Applied. As a result, drift in upstream Templates, the deployment ConfigMap inputs, or rendered child resources (Deployment, Service, Secret, RBAC, HTTPRoute, ReferenceGrant) is not self-healed until the spec generation changes. A future change should add `Owns()` watches on rendered kinds and a Template→Deployment fan-out so drift triggers a reconcile. This is the same drift-watch concern behind HOL-1106 and is the natural follow-up to HOL-1101.
- [IMPORTANT] `internal/deploymentrender/apply.go` orphan cleanup discovery ignores all list errors instead of distinguishing missing-optional-GVR (expected) from RBAC denial / transient API failures (should surface). Pre-existing in the apply package; not introduced by this PR. Worth its own hardening pass.
- [CRITICAL] `config/holos-console/rbac/role.yaml` grants the console service account cluster-wide verbs on `secrets`, RBAC kinds, and the rendered workload kinds. A template author can target any namespace because the render/apply path applies objects to the namespace each manifest declares. This widens the existing ADR 036 cluster-wide footprint and predates HOL-1101 — narrowing it requires either project-namespace allowlisting at apply time or per-namespace RoleBinding fan-out. Out of HOL-1101 scope but should be tracked.

Marking `needs-human-review` on this PR and on HOL-1101. Holding the merge until human triage of the deferred items above.

